PR #66, opened 2014-02-07
Final Confab fixes:

baoilleach:
1. Add docs to confabreport format
2. Make sure that the 'r' option works


wendao:
Hi, I'm trying to use confab in obabel, but it doesn't work as advertised, is it in the master? The example on the webpage just dumps the same sdf.

baoilleach:
Yes. Only in the master.

PR #74, opened 2014-03-27
logP contrib per heavy atom

UnixJunkie:
Would there be a way to integrate this cleanly into Open Babel?


baoilleach:
Will we close this? Any new work should be done on a new PR in any case.

UnixJunkie:
Yes, this can be closed. This task is trivial to do with rdkit nowadays.

PR #85, opened 2014-07-27
Set cmake policy CMP0042 to continue old behaviour

mcs07:
Cmake 3.0.0 gives the following warning when compiling Open Babel on OS X:

```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   openbabel
```

To suppress the warning, policy CMP0042 needs to be set to `OLD` or `NEW` in CMakeLists.txt. This commit simply sets CMP0042 to `OLD`, to continue the pre-3.0 behaviour.

From reading the cmake documentation, it looks like setting the policy to `NEW` _should_ be fine, because CMakeLists.txt defines `CMAKE_INSTALL_NAME_DIR`, which acts as a manual override to any of the changes in default behaviour caused by the new policy. However, I think sticking with `OLD` for now would be best to maintain backwards compatibility with older versions of Cmake (just like what is already done with CMP0005).

More details here:

http://www.cmake.org/cmake/help/v3.0/policy/CMP0042.html
http://www.cmake.org/Wiki/CMake_RPATH_handling
http://www.cmake.org/Wiki/CMake/Policies
http://www.cmake.org/cmake/help/v3.0/prop_tgt/INSTALL_NAME_DIR.html
http://www.cmake.org/cmake/help/v3.0/prop_tgt/MACOSX_RPATH.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,9 @@ set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS 1)
 if(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
   cmake_policy(SET CMP0005 OLD) # add_definitions need updating to set to NEW
-  cmake_policy(SET CMP0042 OLD)
+  if(POLICY CMP0042)
+    cmake_policy(SET CMP0042 OLD)
+  endif()
 endif()
 
 include (CheckCXXCompilerFlag)


baoilleach:
The newest CMakes are now complaining about the use of OLD for this property. Presumably if this is set to NEW, and the Python bindings and binaries still work from a local install, there isn't a problem. Not that I rightly understand what the problem is :-)

baoilleach:
On the other hand, it may require a bump in version required to 2.8.10. I'm not sure whether this is a change we need to do for another few years.

PR #95, opened 2014-10-13
CIFFormat::ReadMolecule now defaults to mmCIFFormat::ReadMolecule

patrickfuller:
Uses mmCIF if found, defaults to existing code otherwise

### DIFF
--- a/src/formats/cifformat.cpp
+++ b/src/formats/cifformat.cpp
@@ -1381,6 +1381,11 @@ namespace OpenBabel
   /////////////////////////////////////////////////////////////////
   bool CIFFormat::ReadMolecule(OBBase* pOb, OBConversion* pConv)
   {
+    // If installed, use the mmCIF parser to read CIF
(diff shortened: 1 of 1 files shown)


kartlee:
Hi, I see an issue with this commit where the atomic number of the element is not retrieved when the symbol has the charge. As a result, we get the following warnings -

Open Babel Warning in GetAtomicNum
Cannot Understand the element label Na1+
.......
Open Babel Warning in GetAtomicNum
Cannot Understand the element label Cl-

The earlier code in cifformat.cpp strip out charge from the symbol to get the atomic number. Is there any reason why this was left out?

patrickfuller:
No reason - I've just never come across CIFs like that. A short-term fix for your CIFs would be to ensure that you're using the `_atom_site_type_symbol` to specify an element.

I'll also aim to make a pull request with this fix in the coming days.

kartlee:
I just created a pull request to openbabel repo - https://github.com/openbabel/openbabel/pull/130. Let me know if it looks okay.

PR #112, opened 2014-11-24
Read and write UCSF Dock docking scores from comments

mwojcikowski:
An option to parse comments provided by UCSF Dock and Chimera. Mol2 writer is also capable to write properties from other sources, like SDF.

Example of such comments:
```
##########	Name:	ZINC04087362
##########	HA_RMSDs:	10.5487
##########	HA_RMSDh:	10.5466
##########	HA_RMSDm:	5.8751
##########	Cluster Size:	19
##########	Grid Score:	-45.986282
##########	Grid_vdw:	-38.770088
##########	Grid_es:	-7.216196
##########	Int_energy:	9.837175

@<TRIPOS>MOLECULE
ZINC04087362
....
```

Note:
There is a bug, although I'm not proficient enough in C to fix that - if N molecules are skipped using "-l" option, then comments for first molecule are lost. I guess MOL2Format::SkipObjects needs to check if line "m-2" has # at the beginning, and then skip few lines back (until hashed lines finish).

### DIFF
--- a/src/formats/mol2format.cpp
+++ b/src/formats/mol2format.cpp
@@ -33,6 +33,8 @@ namespace OpenBabel
       OBConversion::RegisterFormat("mol2",this, "chemical/x-mol2");
       OBConversion::RegisterFormat("ml2",this);
       OBConversion::RegisterFormat("sy2",this);
+      OBConversion::RegisterOptionParam("c", NULL, 0, OBConversion::INOPTIONS);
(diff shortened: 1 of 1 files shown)


ghutchis:
Can you add an "issue" on GitHub to track the bug - that the reader needs to skip back if it sees '#' lines?

mwojcikowski:
Sure, unfortunately "issues" are probably visible for organisation only.

PR #119, opened 2015-01-28
Correct name of inchi library

torcolvin:
This matches the name in src/formats/...


baoilleach:
Does this fix some particular problem that you encountered? I ask because it introduces one, namely it now overwrites the value of a user-specified INCHI_LIBRARY.

PR #120, opened 2015-01-28
Remove useless line, since NAN doesn't exist on msvc

torcolvin:
This fixes a compilation error for msvc. There's not really a consistent way that I see to define NAN in a cross platform way anyway.

### DIFF
--- a/src/formats/mdffformat.cpp
+++ b/src/formats/mdffformat.cpp
@@ -458,9 +458,6 @@ namespace OpenBabel {
       
       if(charge_smb.find(smb) == charge_smb.end() )
         charge_smb[smb] = it->second->GetFormalCharge();
-      else
-        if(charge_smb[smb] != it->second->GetFormalCharge())
-          charge_smb[smb] = NAN;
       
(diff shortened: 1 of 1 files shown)


ghutchis:
I'm obviously OK with fixing a compile error, but I'd prefer to keep in the defensive "else" statement here. Got a different idea? Thanks

ghutchis:
How about numeric_limits::quiet_NaN()

ghutchis:
Bump. Can you please change to use quiet_NaN()?

baoilleach:
Hmmm...I don't have any compile error with MSVC. Maybe this has already been fixed/changed?

baoilleach:
Works for me. Compiles fine with MSVC 12 (2013). Presumably the OP is using an older compiler. NAN is defined as;
 #define NAN        ((float)(INFINITY * 0.0F))

We don't currently have a policy of support older compilers on Windows. I suggest we don't merge unless the OP responds.

baoilleach:
Closing pending further comment....

PR #126, opened 2015-02-13
Check for gz in extension and compress output if necessary

mwojcikowski:
As simple as in title.

I was thinking of implementing other GENOPTIONS, like -f, -l in pybel. I think it would be better and more user-friendly when pybel.readfile() would have explicit arguments: pybel.readfile(....., first=None, last=None). Ideas?


ghutchis:
I think other GENOPTIONS would be great and agree with the idea to have explicit arguments in pebble.readfile().

PR #130, opened 2015-03-06
Separate charge from symbol to derive the atomic number

kartlee:
This was left out by the earlier commit while using mmcif as default
plugin for cif files.


kartlee:
Addressed review comments as part of this hash - 549b182.

PR #137, opened 2015-04-03
Swig fix for python package

mcs07:
This commit updates `scripts/python/openbabel-python.i` to match recent fixes to `scripts/openbabel-python.i`. It also bumps the version number in `scripts/python/setup.py` to prepare for uploading the fixes to PyPI to fix installing the python bindings through `pip install openbabel`.

### DIFF
--- a/scripts/python/openbabel-python.i
+++ b/scripts/python/openbabel-python.i
@@ -229,7 +229,7 @@ CAST_GENERICDATA_TO(SquarePlanarStereo)
 %include <openbabel/math/transform3d.h>
 %include <openbabel/math/spacegroup.h>
 
-# CloneData should be used instead of the following method
+%# CloneData should be used instead of the following method
 %ignore OpenBabel::OBBase::SetData;
 %include <openbabel/base.h>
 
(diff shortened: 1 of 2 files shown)


baoilleach:
Um....remind me why we needed two openbabel-python.i files?

mcs07:
OK I just looked back in the history, and I think the intention was to add `scripts/python/openbabel-python.i` as a symbolic link to the pre-existing `scripts/openbabel-python.i`. This was because package that is submitted to PyPI is created from the contents of the `python` directory. See here: https://github.com/openbabel/openbabel/commit/3cf839f34eb3375f2cebc57193705b8f1192e676

However, it doesn't look like this has worked correctly, and we now have two `openbabel-python.i` files that are out of sync...

It's probably best to figure out a way to go back to having just a single file, but still maintaining the ability to build the python bindings package for submission to PyPI. It may be possible to update `MANIFEST.in` to pull in the file from the parent directory, or if not, maybe we should just keep `scripts/python/openbabel-python.i`, however, that will be inconsistent with the other language bindings.

mcs07:
Hmmm, looks like it was this commit https://github.com/openbabel/openbabel/commit/81caf8f8fd1cb45018f5d84dff100869d3f6e44b that changed it from a symbolic link to a second copy of the file. Not sure what the motivation was.

baoilleach:
I agree - if one file and no symlinks are possible, that would be simplest.

PR #140, opened 2015-04-10
Remove duplicate swig files

mcs07:
Due to some problems with symbolic links, there are currently outdated duplicates of `openbabel-python.i` and `stereo.i` in the `scripts/python` directory. This commit removes these duplicates, but should preserve the ability to install the python bindings independently using `pip`:

- Remove the duplicate swig interface files from the `python` directory.
- Remove the rule specifying those files from `MANIFEST.in`. Unfortunately, we can’t specify files from the parent directory here, so instead:
- Modify `setup.py` to pull in the interface files from the parent directory when preparing the distrubution using `sdist`.

Also:

- Add the rule `exclude openbabel.py` to `MANIFEST.in`, to help prevent accidentally including a copy of the resulting swig bindings in the source distribution.
- Bump the python bindings version number to 1.8.2 to allow these changes to be pushed to PyPI.


mcs07:
@ghutchis / @baoilleach — Are you happy for me to push these changes to PyPI? Someone just needs to run `python setup.py register`, `python setup.py sdist upload` from the python directory.

baoilleach:
Sure.

PR #141, opened 2015-04-18
Fixed an error with the allExplicit option

fredrikw:
Fixed an error with the allExplicit option where the function was opposite to what was stated. Fixed by removing the negation in test.
See line note in  https://github.com/openbabel/openbabel/commit/ba397aa5a01cd3753072fadf9535d0092dfcf784#diff-2a365bc04bfc2f8bea760bcade7466a0


jeffjanes:
Thank you.  I've had half-written email about this bug for a few weeks now, but since I didn't know what the patch was intended to do, I wasn't sure how to describe the problem.  I've tested that this fixes it my issue, and it does.

Issue #143, opened 2015-04-19
Fix policy warnings regarding UsePythonTest with newest cmake


baoilleach:
Running cmake with Python bindings enabled gives lots of policy warnings. 

I've already fixed this locally...just need to check it in.

baoilleach:
Fixed by #146.

PR #146, opened 2015-04-20
Fix for https://github.com/openbabel/openbabel/issues/143

baoilleach:
Tweak to silence CMake warnings about UsePythonTest.cmake (if using recent versions of CMake).


mwojcikowski:
Note, that you can also use github IDs in commits instead of full links, they will be replaced. So in the title you could use #143.

baoilleach:
Thanks mwojcikowski. Funny thing is I did that in my last pull request and it didn't seem to work. Maybe it works fine in the "body" but not in the title. I dunno.

mwojcikowski:
Weird. I've done it in CLI git and discovered it later on GitHub. Anyhow, lets merge it.

PR #148, opened 2015-04-21
Memleak fix

dkoes:


dkoes:
Hold off merging this.. it looks like it is buggy (despite passing make test)

dkoes:
Okay, this isn't a valid fix.  There is no simple fix.  If the istream belongs to the user, you have no way of knowing if it has been deallocated or not by the time you get to the OBConversion destructor (if they were both declared as local variables, for instance, and go out of scope at the same time).  It it is still valid, then the rdbuf needs to be replaced before deallocating pLineEndBuf. It is no longer valid, you can't do this.

The correct solution is to install all filtering on top of the user provided stream.  This could be done relatively easily as a general framework that would apply to the zlib filters (which I'm guessing, from looking at the code, also leak memory).

I'd be willing to take the time to implement this, but would be much more enthusiastic about it if I could use the boost iostreams libraries.  Is this acceptable?

PR #156, opened 2015-04-28
New plugin for generating VDW surface points

mj-harvey:
Hi,

This plugin, "pointcloud" generates points on the vdw surfaces around a molecule
 Radius multipliers and point densities can be specified
 Output is in (pseudo) xyz format

### DIFF
--- a/include/openbabel/plugin.h
+++ b/include/openbabel/plugin.h
@@ -567,6 +567,7 @@ public:
   // charges
   OB_STATIC_PLUGIN(GasteigerCharges, theGasteigerCharges)
   OB_STATIC_PLUGIN(MMFF94Charges, theMMFF94Charges)
+  OB_STATIC_PLUGIN(NoCharges, theNoCharges)
 #ifdef HAVE_EIGEN
   OB_STATIC_PLUGIN(QEqCharges, theQEqCharges)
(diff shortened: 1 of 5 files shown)


ghutchis:
Looks nice, although I have some questions beyond point clouds.. Many people would like solvent-accessible or solvent-excluded surfaces at variable probe radii. The "radius multiplier" seems like it would be more useful from a user perspective as an additive factor (e.g., VdW + 0, +1 + 1.5Ang, etc.)

ghutchis:
Oh, I also forgot to mention that while point clouds are nice, exporting to something like STL (i.e., taking the points and triangulating) would be really highly appreciated.

mj-harvey:
I can add a probe radius too, no problem.

mj-harvey:
I appreciate that would be useful, but it goes beyond our use case. I'm not
sure how to do either the triangulation or write STL, so it will probably
require more resource than I can spare right now. I'll put it on the Todo
list though.

Matt

ghutchis:
The probe radius would be great. No worries about STL / triangulation. I suspect I can coerce an undergrad to work on that this summer.

Two spaces, no tab:
(setq-default indent-tabs-mode nil)
(setq c-basic-offset 2)
(setq default-tab-width 2)

Thanks!

mj-harvey:
closed to resubmit again

PR #159, opened 2015-04-29
Reintroduce the 'charges none' method

mj-harvey:
Hi,

The "charges none" method seems to have been removed at some point.
I'd like to re-introduce it, as we have a dependency on it.

Matt


baoilleach:
Can you go back in time, and see whether there was some reason it was removed? Note that it may be removed accidentally again at some point in the future if you don't add a unit test (hint hint).

ghutchis:
I think it was borked when someone changed the static compiles. I use the normal build path, so I didn't detect this. Probably worth setting up some buildbots.

PR #161, opened 2015-04-29
New plugin: STL output of VDW surface

mj-harvey:
Hidden surfaces not culled, but Makerware seems happy with it.
No control over # of triangles
CPK colours optional 

I'm making this a separate plugin to pointcloud, as it works in quite a different way, and I expect will attract more features over time (eg adding bonds for CPK/molymod style prints)


ghutchis:
This is great, thanks!

PR #167, opened 2015-04-30
Add partialcharge method "fromfile"

ghutchis:
Applies per-atomtype/element charges specified in a file, eg:
babel --partialcharge fromfile:mapping.txt -i mol2 in.mol2 -o mol2 out.mol2

$ cat mapping.txt
C1   1.0
C2   0.5
O1  -1.0
etc


ghutchis:
This is solved by #209

PR #175, opened 2015-05-03
Edit ifdef MSVC to be MSVC version specific

baoilleach:
Fix MSVC2013 build problems as the erfc/erf are now available and this causes the compiler some confusion.


baoilleach:
Oh drat. My brain must be going mushy.

PR #181, opened 2015-05-10
Fix for DeleteHydrogens behaviour #178

### LINKED ISSUE #178: Fix/adjust DeleteHydrogens
Impliciting/expliciting hydrogens should not change the structure but DeleteHydrogens currently removes hydrogens with a specified isotope - it should not.

Unrelated to this, but there's a performance issue with DeleteHydrogens. It should not delegate to DeleteHydrogen but do not whatever needs doing itself.

baoilleach:
Fixes #178


baoilleach:
Actually, there's a bit more to go, so don't merge this yet.

Issue #184, opened 2015-05-26
Make use of hardware bit counting


jeffjanes:
I have a branch "mpopcnt" which adds the -mpopcnt flag to the compilation.  This uses hardware support for the "__builtin_popcount" operations.  

This makes searches in fast-search indexes 3 to 6 times faster.

I know little about cmake, and I did this through random mutagenesis.  Does anyone know what would be the correct way to integrate this into the build script?

baoilleach:
Many types of awesome! I thought this was already happening but apparently not.

I'll have a look as soon as possible if no-one else gets there first.

baoilleach:
I think the correct approach would be to add a Check script that tries to compile and run a one-line cpp file containing __builtin_popcount with -mpopcnt. This fails with "illegal instruction" on systems that don't support it. Hopefully all of this is possible in practice.

jeffjanes:
Do we have to worry about cross-compilation where the target platform might not support it even though the compiling one does?  Or is that not supported in the first place?

ghutchis:
I believe Cmake handles those cases - that is, the various compiler checks would be for the cross-compiler.

cowsandmilk:
(i) checking if the compiler has support for a flag using CMake:
check_c_compiler_flag(-mpopcnt HAS_POPCOUNT)

(ii) I think the check Noel is referring to, and @jeffjanes for cross-compilation, is processor support. I agree with @jeffjanes that there is an issue where the library could be compiled on a computer with processor support, then run on a computer without it. The most common example is packaging for operating systems. Shipping rpms that require the popcnt instruction would be less than ideal.

I also know of several pharma companies that use a module system such that openbabel is run from the same binaries/dynamic libraries on disparate hardware from new to 6+ years old.

So, I think there should be some CMake variable set by the user at compilation time for turning it on or off. I don't have strong opinions on defaulting to on or off.

jeffjanes:
Perhaps we could compile both versions of Tanimoto, and then at run time when it opens a fs file, make a dummy __builtin_popcount call and trap the illegal instruction signal (if it is trappable) and change a function pointer to use the other version.  Of course trying to do that in a portable way would also be problematic.  Another the issue is that the "emulated" __builtin_popcount you get when lacking the -mpopcnt flag is probably better than the fallback which openbabel uses.  So we would need to improve that fall-back method as well, or some systems would end up worse off.  

[...]

PR #186, opened 2015-05-30
Add support for hardware popcount

baoilleach:
Add -march=native option. Can be turned off by -DOPTIMIZE_NATIVE=OFF.

See issue #184

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,7 +105,6 @@ if(GLIBC_24_COMPATIBLE AND UNIX)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--wrap=memcpy")
 endif()
 
-
 option(BUILD_SHARED "enable shared build support" ON)
 option(BUILD_MIXED "enable linux relocatable binaries support" OFF)
 option(WITH_STATIC_INCHI "enable static inchi and related plugins" OFF)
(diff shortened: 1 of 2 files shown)


jeffjanes:
I've tested this on Linux and it works well for me.  The hardware support is correctly used on a more modern hardware and correctly not attempted on the older one.  Thanks.

As you noted, it fails to make use of the hardware popcount support on VirtualBox instances, but if compiled with -mpopcnt flag anyway it will actually use it.  Should an option be added for turning this on in cases such that as that?  Something like "-DFORCE_POPCNT=ON".

I don't know how common VirtualBox is, but apparently both of us use it so it can't be too rare.

VMWare instances work correctly.  So does whatever Amazon uses for their t2.small Amazon Linux instances.

baoilleach:
Actually Roger (a former gcc developer) looked into the Virtual Box issue, and he reckons there's a bug already filed, and that it has been fixed in the virtualisation technology that virtual box uses (qemu or one of those). So I'm reading this as something that will be fixed by-and-by when Virtual Box updates to the latest qemu, rather than something we need to address.

However, OB should be making it easier to set arbitary flags. It should be possible to set CMAKE_CXX_FLAGS at the command-line and have this passed through to the build. Right now, we trample over any user-defined flags. But one thing at a time...

baoilleach:
FYI, VirtualBox 5.0  - now in beta - has popcnt (https://forums.virtualbox.org/viewtopic.php?f=1&t=66905)

PR #187, opened 2015-05-31
Don't set precision to 0 when reporting time estimate

jeffjanes:
Changing the stream precision to 0 seems like a bad idea.

I'd rather see a reported time of 90.5956 minutes and
mentally round it to the number of sig figs I think is
appropriate, than see 1e+02 minutes and be left to wonder
where between 50.1 and 149.9 that might mean.


baoilleach:
This is definitely an improvement on the current output, but how about outputting so many days+hours, or minutes+seconds. Fair enough if you don't have time, but just a thought...

PR #188, opened 2015-06-03
New option in VASP format. Custom atom sorting.

orex:
The option allows to sort atom not alphabetically, but using the custom list. For example, an option -xw "Al Si O" force obabel to put Al atoms in a first in the output POSCAR file, next Si and in the end O atoms. The option is really useful if you have POTCAR file is already exists. The order of elements in both POSCAR and POTCAR files should be the same. The option is fully compatible with previous alphabetical implementation.

### DIFF
--- a/src/formats/vaspformat.cpp
+++ b/src/formats/vaspformat.cpp
@@ -188,6 +188,7 @@ namespace OpenBabel {
     // Build unit cell
     OBUnitCell *cell = new OBUnitCell;
     cell->SetData(x_vec, y_vec, z_vec);
+    cell->SetSpaceGroup(1);
     pmol->SetData(cell);
 
     // Next comes either a list of numbers that represent the stoichiometry of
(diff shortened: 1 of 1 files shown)


baoilleach:
Hi Orex, while this is in theory compatible with the existing 'w' option, we don't support options with an optional parameter. This will have to be moved to a new option, and the Documentation() string for VASP updated. Is this something you might have time to sort out?

orex:
I do have time to manage your request, but I don't think, that I understand you fully. Can you write a detail task.
1) The issue of the option is that you can specify 0,1,2,3..N atoms names in the argument. The atoms, which you don't specify are optional. Ex. -xw  "Al" put Al first, the rest atoms by default etc.
2) If you want me to rename the option, let it be -xy, what should I do with command "-xw -xy 'Al' "?

baoilleach:
Great.

The ultimate problem is that the openbabel GUI supports either boolean options (uses a checkbox)  or options with parameters (uses a textbox), but not a combination of the two. So the GUI right now cannot provide the user a way to choose the three combinations (no "xw", "xw" with no text, "xw" with text).

But rather than change the GUI, it's clearer to the user if you just add another option.

Regarding (2), it's up to you - I'm not familiar with the use cases. Here is my suggestion:
* no option: default order (presumably this is the order of atoms in the input molecule)
* -xw : Sort atoms by atomic number
* -xy 'Al': Al first, then default order
* -xw -xy 'Al': Al first, then sort atoms by atomic number

PR #191, opened 2015-06-15
OBConversion improvements

dkoes:
Summary:
*fixes memory leak 
*more robust support for gzipped files
*more convenient methods for opening files (format auto-detection)
*fix bug with xml reader when reading files with multiple molecules

I have made every effort to maintain backwards compatibility (even when the existing behaviour was dubious) but there are some unavoidable changes.  Specifically, streams passed to a OBConversion should not be deallocated before the OBConversion is deallocated or the stream is replaced.

Details:
This was all motivated by a memory leak due to the line ending filter.  OBConversion would swap out the rdbuf of the passed input stream with its own filtering rdbuf to normalize all line endings.  This means a user-provided data structure is having its internals messed with and there is no way to properly free the filtering streambuf (since OBConversion doesn't control the stream itself).  This would cause my long running program that opens millions of files to eventually consume all the memory on the machine and crash.  Not acceptable.

The solution is to not mess with the internals of the passed stream, but wrap the stream with a filtering stream. The full stream is wrapped, not just the underlying streambuf, in order to (more or less) keep the state of the user-provided stream consistent.  That is, when OBConversion reads to the end of the file, the good bit of the user-provided steram will also be turned off (some test code depended on this behavior).

GetInStream/GetOutStream may now return a stream that is not identical to the user-provided stream (since it's the wrapped version).  In addition to the newline filter, streams may be wrapped if compression is enabled. 

Compression is now provided by virtue of wrapping the stream (and leaving it that way until all writing/reading is done).  GetInStream/GetOutStream provide this wrapped stream and so can be read as if the underlying stream were not compressed.

FormatFromExt can take a second argument, a bool passed by reference, that will be set to true if the file is gzipped.  For example, "sdf.gz" will return the SDF format and set the bool to true.

SetInAndOutFormat, SetInFormat, SetOutFormat have all been modified to take an optional isgzip parameter which, if true, enables compression.   The functions GetInGzipped/GetOutGzipped are also provided.  Essentially, the compression state of the file is part of its format (but has to be handled separately due to how OBFormats work).

So to get and set the full format of a file, including compression state:
bool isgzipped;
OBFormat *format = FormatFromExt(fname, isgzipped);
conv.SetInFormat(format, isgzipped);

Convert(in,out) has different semantics.  It no longer sets the input/output streams with the passed arguments.  This is because existing code (e.g. babel) was deallocating these provided streams before the OBConversion object they were passed to.  Since Convert consumes all its input stream, it doesn't seem unreasonable for it not to save the streams.  By not keeping pointers to the provided streams, the possibility of a use after free error goes away.  The constructor, SetInStream/SetOutStream, ReadFile/WriteFile, and Read/Write functions can still be used to set persistent streams.

Internally, SetInStream/SetOutStream are used to setup/teardown streams, as opposed to the previous approach which was to duplicate (slightly different) blocks of code everywhere this happened (for instance, to check for gzip, set NeedToFree, etc). OBConversion keeps track of what streams it owns the memory to.  SetInStream(NULL) or SetOutStream(NULL) will take care of cleaning up the current stream.

The semantics of the copy constructor have been refined and an assignment operator added.  Previously the state was just copied resulting in two objects that both thought they owned the same memory and had to delete it.  This only worked because of the implicit assumption that OBConversions were only copied to create AuxConv objects for XML parsing (why these need to be subclasses, I know not) and there's special logic to avoid duplicate deletions in this case.  Now the semantics are that only the originating OBConversion retains ownership of any memory it allocated.  This eliminates the possibility of a free-after-free, but not a use-after-free (the copy constructor just shouldn't be used).

ReadFile/WriteFile will now extract the format from the filename if the format is not already set.  This includes setting the gzip status of the file. I've also added a convenience constructor to OBConversion that takes filenames.  This removes a lot of the boilerplate required for doing simple tasks.  For example:

OBMol mol;
OBConversion conv(infile);
while(conv.Read(&mol)) {
 //do stuff
}

Gzip support has been improved to work properly on files that are concatenations of gzip files (this is suppose to be a valid thing to do).

The XML reader would return true to ReadMolecule even if it hadn't read a molecule in a multi-molecule file (because it was absorbing the ending `</cml>` tag).  Now it returns false.

---
I suspect these changes will cause some people's code to crash.  However, the cause of the crash should be immediately identifiable - the lifetime of the OBConversion object is longer than the lifetime of the input stream passed to it (and the OBConversion destructor crashes trying to flush to the no longer available user-provided stream).  It should not be surprising to anyone that leaving a stale pointer to a stream in OBConversion could cause bad things to happen and the fix is easy (call SetInStream(NULL) when done reading the input stream).  

On the flip side, leaving the current memory leak in results in a much less obvious and more difficult to debug crash (and it isn't the developer's fault).  I think this change is worth the minimal breakage of bad code that it will entail.


mwojcikowski:
Just one minor comment: why introduce additional argument for FormatFromExt? Wouldn't a check if recognized format is "gz" and if so returning actual format be sufficient?

Other than that I'm all in, since the memory leak was an issue in one of mine projects.

dkoes:
How are you suppose to auto-recognize an output file that hasn't been 
written yet as gzip? I think of the compression as part of the format - 
you shouldn't need an open stream in order to fully specify what it 
should be. An example of something that now works the way you would expect:

babel x.mol2 x.sdf.gz  # previously x.sdf.gz was text output

In my perfect world compression state would be embedded within the 
OBFormat class, but since these are singletons that doesn't work.

David Koes
Assistant Professor
Computational and Systems Biology
University of Pittsburgh

dkoes:
Fixed windows build errors which were pointing to serious problems that should have at least generated a warning with gcc.

mwojcikowski:
I thought of using old GENOPTIONS "z" interface internally to specify that output file should be compresse, but I've just noticed you got rid of it.

dkoes:
For backwards compatibility, that option is still there.  However, if 
the option is set after a stream is set, then it doesn't have an effect.
For example

OBConversion conv(&infile, &outfile);
conv.SetOptions("z", OBConversion::GENOPTIONS);

No longer works, but the following works for both new and old code:

     OBConversion conv;
     OBFormat *format = conv.FormatFromExt(fname);
     conv.SetOptions("z", OBConversion::GENOPTIONS);
     conv.SetInStream(&infile);
     conv.SetOutStream(&outfile);
     conv.SetInAndOutFormats(format,format);
     OBMol mol;

     while(conv.Read(&mol))
     {
         conv.Write(&mol);
     }

The big difference is the new code will generate a significantly more 
compressed file since it isn't creating gzip headers and footers for 
each molecule and is compressing the full output as opposed to each 
individual molecule.  We're talking 35% smaller.

I can make setting the 'z' option convert any existing non-gzipped 
streams to gzipped streams if desired.

Thanks,

David Koes
Assistant Professor
Computational and Systems Biology
University of Pittsburgh

mwojcikowski:
I can report that reading gzipped file via pybel is broken in this PR.
It could be easily fixed by setting 'z' in GENOPTION [https://github.com/openbabel/openbabel/blob/master/scripts/python/pybel.py#L253], or is there a better solution?

dkoes:
pybel should be fixed now.

mwojcikowski:
I can confirm - pybel works as expected. Finally the memory leak is also gone in Python. I made a couple of runs in ipython through ZINC and there is no sign of it any more! Cool :+1:

mwojcikowski:
I just mentioned, you are using the old version of CLI
> babel x.mol2 x.sdf.gz  # previously x.sdf.gz was text output

Have you checked obabel? For me it's not working, meaning
> obabel x.mol2 -O x.sdf.gz

is still not compressed...

PR #199, opened 2015-07-10
Added further checks for zlib availability in OBConversion

philthiel:
More checks were necessary to make all occurrences
of zlib-dependent code build conditional.


ghutchis:
Thanks!

PR #201, opened 2015-07-15
do not blow away multiple conformers when adding hydrogens

dkoes:
I previous "fix" several years ago added successive calls to BeginModify/EndModify in AddHydrogens to ensure that coordinates were set (PR#1665519). Unfortunately, this completely blows away all but the set conformer.  This, despite the AddHydrogen code correctly handling the multiple conformer cases (at least it looks like it).

I believe that all that is needed to resolve the original bug is to make sure coordinates are set on atoms.  The original bug report claims test cases were added for the bug and this change doesn't result in any test failures, so I assume it fixes the original bug while retaining multiple conformers.


dkoes:
Any suggestions on how to debug the appveyor build failure? The test passes for me..

baoilleach:
I hit rebuild. Let's see.

baoilleach:
Looks legit. When you say the test passes for you, are you matching the appveyor build (i.e. same OS and compiler version)?

dkoes:
No, I'm on linux.  The test passes, but when I run it through valgrind there are all sorts of errors.  I get the same errors even without the fix.  I'll look into it a little.

Might be interesting to run at least a subset of the tests through valgrind as part of the autovalidation process..

dkoes:
Actually, the valgrind failures were with invalidsmiles, not invalidsmarts, so I have no idea what the problem might be.  In my checkout I have more tests, is it possible the auto-tester is running something different for this test?

dkoes:
Commit b3920bc undid the initial change and appveyor got the same failure.  I think this failure is a reason not to accept this change.

baoilleach:
I'm looking into the problem with the original commit.

baoilleach:
I'm not able to reproduce it either. I'm turning on verbose tests and rebuilding the original commit on Appveyor...

baoilleach:
Just to doublecheck that nothing else is going on, can you checkout master, make a whitespace change, and send a pull request. If this passes, can you apply the original commit to that.

baoilleach:
Drat. I don't understand. Right now, I can only assume it's a quirk of the Appveyor infrastructure, and I do note that they recently changed something or other. Of course, it could be a real bug, but even if so, it's not related to your patch and we can't reproduce it. Apologies for the inconvenience. I'll tell Appveyor to skip that particular test and rebuild.

PR #214, opened 2015-08-03
Nwchem improvements

ghutchis:
Fixes:

Fixed detection of atomic number when coordinates in input file were set as z-matrix
Fixed detection of imaginary frequencies
New parsers of calculations:

ZTS calculation
OBElectronicTransferData extraction from TDDFT calculations
Single point energy calculation
New parsers of calculation details :

Partial charges
Orbital data
Energies
Optimization steps
Dipole moments
Restructurization of code:

Each parser in its own method
String constants replaced by their mnemonic names

### DIFF
--- a/src/formats/nwchemformat.cpp
+++ b/src/formats/nwchemformat.cpp
@@ -16,6 +16,13 @@ GNU General Public License for more details.
 
 #include <openbabel/obmolecformat.h>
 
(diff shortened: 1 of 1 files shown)


ghutchis:
Replaces #213 - only modifies NWChem parser

PR #217, opened 2015-08-15
Fix swig include_dirs+library_dirs in setup.py

mcs07:
When installing the python bindings independently using `pip` or `python setup.py install`, distutils/setuptools now appear to add some default SWIG `include_dirs` and `library_dirs`, so we should append the openbabel paths to these lists rather than overwriting them entirely.


ghutchis:
@baoilleach @patrickfuller Any comments on this one?

baoilleach:
Use of that file is not documented so anything goes.

patrickfuller:
No comments here, but the change makes sense.

PR #231, opened 2015-09-22
small changes to compile on windows with msvc

lenkd:


baoilleach:
Thanks for the patch. To make reviewing easier, in future could you provide more descriptive commits and have separate pull requests (e.g. on different branches) for unrelated changes.

PR #234, opened 2015-10-16
fix for segfaulting similarity search with fastsearch index

dkoes:
This fixes a segfault when trying to use a fastsearch index
    babel mymols.fs -ifs -sN#Cc1ccccc1C#N results.smi
would segfault since the fastsearch plugin assumed that the -s argument was set and was indexing into an empty vector.  If it encounters this situation it will now generate a message instead.

More importantly, the reason the argument was missing is because plugin parameter options were getting registered *after* the parameters were parsed in babel.cpp.  Specifically, in DoOption in babel.cpp `nParams` for -s was zero since its number of arguments hadn't been registered yet. This is due to the commit 3cb13fb2155740ad632c0cd49fcbf9686c78f2ee.  This made plugin loading something that happens greedily when the first plugin is retrieved, whereas before it happened in the OBConversion constructor.

The provided fix makes LoadAllPlugins a public function and calls it explicitly in babel.cpp before parsing commandline arguments.

### DIFF
--- a/include/openbabel/plugin.h
+++ b/include/openbabel/plugin.h
@@ -124,6 +124,9 @@ public:
   ///Returns the map of the subtypes
   virtual PluginMapType& GetMap() const =0;
 
+  ///Load all plugins (formats, fingerprints, forcefields etc.)
+  static void LoadAllPlugins();
+
 protected:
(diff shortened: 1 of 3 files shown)


ghutchis:
D'oh!

PR #241, opened 2015-12-14
Fix infinite loop in eem calculation

aandi:
In a previous commit, the type of the variable i was changed from int to unsigned int. As a result, the back substitution loop became infinite (comparison of unsigned int with zero) and the arrays B and A were accessed beyond their bounds causing a segfault.

### DIFF
--- a/src/charges/eem.cpp
+++ b/src/charges/eem.cpp
@@ -138,7 +138,7 @@ namespace OpenBabel
 
       // Search for parameters for a particular atom type
       bool found = false;
-      for(int j = 0; j < _parameters.size(); j++) {
+      for(unsigned int j = 0; j < _parameters.size(); j++) {
         if((_parameters[j].Z == n && _parameters[j].bond_order == b) ||
             (_parameters[j].Z == n && _parameters[j].bond_order == - 1) ||
             (_parameters[j].Z == -1 && _parameters[j].bond_order == -1)) {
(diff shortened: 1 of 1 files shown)


cowsandmilk:
I'm pretty sure your code currently is reading off the end of the buffers by reading/modifying B[dim] and reading A[dim][dim]; and your code skips the i=0 case.

Basically, your code doesn't match the behavior of the old code using signed integers.

You could use a do { } while or just decrement i as the first line of the for loop.

aandi:
Thanks for the response. It works as it is supposed to, but it is a bit harder to read. I put the decrement as a postfix into the loop condition. The condition is evaluated at the beginning of each iteration and i is decreased right after that. So when the condition is valid for i = 1, i is subsequently set to zero (i=0 case). Hence the loop has to start at dim and not dim - 1.

Minimal example:
```c++
#include <iostream>
int main() {
  // correct signed code
  int i, dim = 3;
  for (i = dim - 1; i >= 0; --i)
    std::cout << i;
  // proposed fix
  unsigned int j, dim2 = 3;
  for (j = dim2; j-- != 0; )
    std::cout << j;
}
```
Output: ```210210```

I guess it is a matter of taste. I prefer the proposed version for its compactness. But of course I can change it to:
```c++
for (i = dim; i != 0; ) 
  {
    --i; 
    // ... 
```
Would you prefer that? I am not sure why the travis build fails though.

ghutchis:
Can this be re-based on the new EEM code? Thanks.

ghutchis:
The travis build failures are unrelated to the patch - updating the patch with current master should fix the build problems.

aandi:
Sure. I re-based the patch. The other EEM patch also addressed the segfault but fixed it by changing the loop variable to an integer. This patch now fixes all warnings due to signedness.

ghutchis:
Thanks, that's great.

PR #243, opened 2015-12-16
Integrate Cheminf 2015 EEM parameters

krab1k:
Extend EEM to use additional parameters from http://www.jcheminf.com/content/7/1/59 and fix some issues along the way.


ghutchis:
If you can fix the documentation mistake, I'm happy to merge this. Looks great, and I'm sure there will be more EEM methods added in the future.

krab1k:
It's fixed now, thank you.

PR #254, opened 2016-02-03
Clean space group name before anything

alexandr-fonari:
This fixes the issue with reading space groups from PDB (and maybe other formats).


ghutchis:
This was another easy one. Thanks.

PR #257, opened 2016-02-09
Inspecting PDB insertion codes

smoe:
Hello,

I am presenting a first patch towards the acceptance of PDB insertion codes.
A more detailed description on what insertion codes are and how they are used I put as a comment to commit https://github.com/openbabel/openbabel/commit/46bb2d1d42b23938c6446680632477e950643e1d - in short, it is a single character that extends the residue ID and may become particularly important when it is only that character that distinguishes two consecutive residues representing the same amino acid.

The patch is complete in the sense that PDB files with insertion codes are now properly read and transformations into other formats, tested was FASTA,  behave correctly, i.e. no residue is omitted. The patch is incomplete in the sense that related formats like PDBQT (and possibly others beyond my immediate expertise) are not yet addressed. Also, I yet did not address any writing of PDB formats. I am prepared to continue this work, just hoping for a bit of a review and guidance in that I am not completely erroneous in how I addressed the insertion codes in the first place.

I am attaching relevant lines of a structure with that feature
[test_insertioncode.pdb.txt](https://github.com/openbabel/openbabel/files/123521/test_insertioncode.pdb.txt)
which transformed looks like
<pre>
$ obabel -ipdb openbabel/test_insertioncode.pdb -ofasta 
>openbabel/test_insertioncode.pdb 6 bp; generated with OpenBabel 2.3.90
SVSSSY
1 molecule converted
</pre>
while the current implementation omits one (the middle) S.


smoe:
obabel -ipdb filename.pdb -opdb
will thanks to https://github.com/openbabel/openbabel/commit/bf60af2d0677f649bfd50dc6ce5b74027c5da6f9 now write the insertion code if the original entry in filename.pdb had it. If there can be options to the reading of PDB entries, then a reasonable thing to do may be to remove those insertion codes and instead just increment an offset for resnum. Those insertion codes are misstreated by a range of PDB-reading (well, mostly reading) tools and OpenBabel may be about the first source for the search of a respective insertioncode elimination.

smoe:
I just cross-checked, there is no residue and hence no insertion code in PDBQT. From my side, this patch is now ready.

ghutchis:
I think the one thing I'd really like is a unit test for this (e.g., an example PDB file and then the FASTA sequence check). That way if someone breaks the insertion codes with a commit, it won't pass the unit tests.

Otherwise, the code looks OK.

smoe:
I found your https://github.com/openbabel/openbabel/tree/master/test directory and will somehow manage :)

smoe:
Works!

PR #260, opened 2016-02-10
Allow only unique symmetry transformations

afonari:
This should match the correct space group when the same transformation is present:

```
_symmetry_equiv_pos_as_xyz
x,y,z
x,y,z
```


afonari:
Thank you for the fast review and merge.

PR #261, opened 2016-02-12
Correctly convert to P1 cell in mmcif format when space group is not found

afonari:
For several cases SpaceGroup::Find returns `NULL`. The idea of this patch is to try to convert to P1 group (apply all available symmetry transformations). This way we might get at least something better than just atoms present in the asymmetric unit with no symmetry operations..


ghutchis:
Can you give me a use case or example file for this one? The code looks fine, but I'm curious about similar cases. For example, would this be able to pull out the molecule when the CIF "wraps" it around the unit cell?

afonari:
> Can you give me a use case or example file for this one?

Check this file out: http://www.crystallography.net/cod/cif/1/01/00/1010022.cif
When running, the result is (one of the paths to get NULL from SpaceGroup::Find):
`Unknown space group error, please file a bug report.`

The reason is a weird group name: `Hermann-Mauguin symmetry space group	R -3 m HR`
With the patch, we get nice P1 file (since the transformations are correct [which is not always the case]).

> would this be able to pull out the molecule when the CIF "wraps" it around the unit cell?

Not sure what you mean. I would guess it means that you have coordinates outside the unit cell:
`C1 1.2 -1.2 1.2`
which can be translated back to:
`C1 0.2 -0.2 0.2`

Looking into fillUC code it says:
```
 // Now loop over all symmetry operations, and generate symmetric atoms one at a time
 // Avoid creating overlapping atoms (duplicate), and bring back atoms within the unit cell
```

So the answer seems, yes.

Let me know if you have any more questions.

PR #269, opened 2016-03-01
Moved storing total charge in gaussformat.cpp

dspoel:
The total charge in an OBMol may be changed by EndModify or
other routines involved with changing bondorder and or atom types.
This patch moves the assignment to the end of the file to make
sure the total charge does not get lost.


dspoel:
Hm, maybe I updated it twice, hope the last one is fine.

PR #273, opened 2016-03-07
The findFiles member should return 0 for error

cryos:
The Windows equivalent returns 0 for error/no files, this variant should
do the same. This member is used to load all plugins, and can result in
an infinite loop when -1 is returned.


cryos:
@ghutchis this fixes the infinite loop issue I was seeing when the BABEL_LIBDIR variable was set to something containing no files. You could of course go the other way and change the code to respect the -1 error introduced by the unix variant, but it seemed better to adjust it to return 0 as the windows variant is. I verified the loop then terminates as expected whether files are found or not.

PR #278, opened 2016-03-14
GAFF atom types

mmghahremanpour:
gaff.prm and bondtyp.txt are updated in order to reproduce the atom types based on the Wang et al. Vol. 25, No. 9 . Journal of computational Chemistry. 2004


mmghahremanpour:
11 energy values have been updated in gaffresults.txt because the old ones were based upon the wrong atom types.

ghutchis:
Is there some way we can add a unit test for the correct GAFF atom types (i.e., beyond the current energy evaluations)? This way, if some future commit accidentally breaks GAFF typing, we would know.

mmghahremanpour:
Actually, what we can do is to select a subset of molecules (e.x 30 molecules) and generate their atom types to be used as the reference. Then, we can add a new test to compare the atom types of these compounds with those used as the reference. Therefore, if someone changes gaff.prm or anything related, the test will fail if the changes cause a mismatch.

PR #280, opened 2016-03-15
Default OPTIMIZE_NATIVE to OFF, fixes #279

### LINKED ISSUE #279: Default OPTIMIZE_NATIVE to OFF - unexpected behavior
I spent quite a bit of time debugging obabel on the Mac, and was seeing issues with an illegal instruction pretty early on as obabel started up. This was a pretty difficult one to debug as everything worked fine on the build machine, but I got a very generic error 14 on my laptop when trying to use the packaged obabel, in the end I found the answer through searching on Google and seeing a similar issue for Docker.

CMake and the upstream compilers don't enable this by default as the binaries can be quite specific for the GPU you are building on. In this case it was a pretty new Mac Pro, and trying to run them on an older Macbook Pro laptop. The comments seem incorrect too, as they have little to do with cross compiling and more to do with optimizations for the specific CPU on the build machine.

I think adding a section on optimization, and advising to turn this on there (with a big warning about the binaries being quite CPU dependent at that stage) would be great, but please consider turning this off by default to avoid having others spend time debugging strange run time errors.

cryos:
Default the OPTIMIZE_NATIVE option to OFF, and update the documentation
to more accurately reflect the optimization for the machines CPU
architecture. This isn't related to cross-compiling, so replaced
references to it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -467,13 +467,13 @@ else(CAIRO_FOUND)
   message(STATUS "Could NOT find Cairo. PNG output will NOT be supported.")
 endif(CAIRO_FOUND)
 
-# Should we enable optimizations for the native architecture?
+# Should we enable optimizations for the native CPU architecture?
 # (this will speed up similarity calculations and maybe those involving Eigen)
 if(NOT MSVC)
-  option(OPTIMIZE_NATIVE "Optimize for native architecture. Turn off if cross-compiling." ON)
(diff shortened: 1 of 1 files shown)


ghutchis:
@baoilleach I'm inclined to merge this. We can add clear documentation in the installation instructions, but I think for most use, it's preferable to have portable binaries than perfect speed.

baoilleach:
Fair enough.

jeffjanes:
Now that the default is off, shouldn't the cmake hint be inverted so that it tells you how to turn it back on, rather than still telling you how to turn it off?

PR #283, opened 2016-03-16
Allow compilation as static binaries

torcolvin:
Some changes need to compile with:

-DBUILD_SHARED=OFF -DWITH_INCHI=ON -DWITH_STATIC_INCHI=ON


ghutchis:
I've added a static build to the Travis CI tests in #284. Hopefully this type of build will be fixed now.

PR #287, opened 2016-03-24
CMakeLists.txt fix for wrongly hard-coded path

arkose:
The main improvement is getting rid of the hard-coded path by using the already-existing variable.

(Also introduces the possibility of the include install path to something else using -DOB_INCLUDE_DIRS in cmake; this was my personal reason for looking through and editing the CMakeLists.txt file, I wanted the option to install headers to ~/include/openbabel/ instead of ~/include/openbabel-2.0/openbabel/, for example.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,9 @@ else() # Windows - bin dir = lib dir to load libraries
     endif(ENABLE_VERSIONED_FORMATS)
   endif()
 endif()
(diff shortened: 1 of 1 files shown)


baoilleach:
Usually it's good to have one feature or fix per pull request.

arkose:
>baoilleach: Usually its good to have one feature or fix per pull request.

To be clear: should I split the patch in this case?

baoilleach:
There's no need. Just something for next time.

PR #295, opened 2016-04-06
thread_local storge for global variables with state

dkoes:
OpenBabel uses a number of global objects for performing various kinds of perception and these objects maintain state (mostly in the form of smarts patterns).  This introduces race conditions in a multithreaded program that can result in segmentation faults.  For example, if two instances of bond perception are called at the same time, one might zero out the match list in the smarts pattern while the other is reading it. 

I have spent many hours trying to work around this probably by guarding calls with locks or parallelizing with processes instead of threads, but since perception routines are called as needed, it can be quite difficult to get right and it really shouldn't be necessary.

Today I realized that making these global variables thread local should solve all these problems (and stop pharmit from crashing).  **This means that openbabel needs to be compiled with -std=cx11.**  This patch should gracefully degrade into the old buggy behaviour if this isn't an option.

Note I may not have caught all the global objects that need this fix, but I think I at least got the ones with smarts patterns.  Global declarations have been consolidated into mol.h and prefixed with a new macro THREAD_LOCAL.


EricTing:
Is there a automatic way to check whether a global objects may cause thread racing?

dkoes:
You could try something like helgrind:
http://valgrind.org/docs/manual/hg-manual.html
You'd need to write a multi-threaded program that exercised all the relevant code though.  Really, any global state that isn't read-only is highly suspect.

dkoes:
The Travis-CImake test with RUN_CTEST='true' fails for me even without the patch.

ghutchis:
Yeah, let me check on that one.

ghutchis:
I've been intrigued by ThreadSanitizer, but don't think it runs on Mac yet:
https://github.com/google/sanitizers/wiki

ghutchis:
I will need to think carefully about this one in terms of 2.4 vs. 2.5. I'd prefer to cut the 2.4 release ASAP and don't know if I have time to think through potential repercussions. Seems like the right change, but from a release manager perspective, I'm likely to push for 2.5.

ghutchis:
Your patch fails on SWIG / building the python bindings - mol.h doesn't define OBBondTyper.

```
/home/travis/build/openbabel/openbabel/include/openbabel/mol.h: In function ‘int Swig_var_bondtyper_set(PyObject*)’:
/home/travis/build/openbabel/openbabel/include/openbabel/mol.h:771:40: error: ‘OpenBabel::bondtyper’ has incomplete type
/home/travis/build/openbabel/openbabel/include/openbabel/mol.h:73:9: error: forward declaration of ‘struct OpenBabel::OBBondTyper’
```

dkoes:
I can't reproduce the clang link failure.  It may be a function of what libraries are installed:
http://stackoverflow.com/questions/29322666/undefined-reference-to-cxa-thread-atexitcxxabi-when-compiling-with-libc

ghutchis:
I'm going to revisit once I cut the 2.4 release. I suspect the test failure is minor.

ghutchis:
Can we revisit this? Now that 2.4 is branched and released, I'd like to merge this. Thanks.

baoilleach:
I see stuff happening on this branch, but I don't think it's related to this pull request. Can this be closed?

dkoes:
Your recent changes should take care of what this was trying to fix... when I get some time I'll try and stress test your changes.

baoilleach:
Right so - I'll close this.

Issue #177, opened 2015-05-04
Fix Clang warnings (includes some bugs)


baoilleach:
See warnings at (for example):
https://travis-ci.org/openbabel/openbabel/jobs/60871323

anishp:
Hi, I'd like to take up this issue. Could you assign this to me?

baoilleach:
Great.

PR #296, opened 2016-04-08
fixed most of the clang warnings

anishp:
partly fixes issue #177


ghutchis:
Wow, thanks, that's helpful. We've mostly ignored the warnings from the inchi code, since that library has really poor code.

I'll do a bit more review, but this looks fine.

PR #297, opened 2016-04-10
Make obtautomer exit with status 0 on success

jeffjanes:
Well-behaved programs are supposed to exit with status 0 upon
successful completion. Make obtautomer follow this convention.

There are other programs which exit with 1 upon successful completion,
such as obprop, obsym, obprobe and perhaps others.  But since they
have been released in that state, maybe it is better not to change
them now.


ghutchis:
I'm fine with all of them doing exit 0 on success. I doubt anyone has been checking the return value or they would have complained previously.

PR #303, opened 2016-05-13
UFF forcefield: fix missing SetupPointers()

aandi:
I noticed slight differences in the FF calculation results for repeated calculations of the same structure using UFF. The error stems from a missing SetupPointers() override.
gaff, ghemical, mmff94 all use OBFFCalculation objects and override the forcefield base function SetupPointers() to assure that the coordinate pointers are setup properly. The patch fixes the error, such that repeated calculations lead to the same results.


ghutchis:
Thanks!

PR #304, opened 2016-05-26
Changes for MSVC2010 compilation.

psavery:


psavery:
I changed

typedef __int8 uint8_t;
typedef __int16 uint16_t;

to

typedef unsigned __int8 uint8_t;
typedef unsigned __int16 uint16_t;

I think that's a better fix

baoilleach:
Can't OPENBABEL_USE_SYSTEM_INCHI be controlled as a CMake option rather
than removing those lines from CMakeLists.txt? I'm guessing that whoever
added that option is not going to be happy when it disappears.

psavery:
Yeah, I think you're right. We should be able to just control that with CMake options. I removed that change.

PR #306, opened 2016-06-02
read additional geometry optimization data

lenkd:


ghutchis:
Can you be a bit more specific about the changes?

lenkd:
Yes, sorry for the short commit infos!
Now after an optimization run every geometry step will be read into the displacement structure to be able to visualize the optimisation steps and the energy changes in future.
Small bugfix to read the IR-spectra correct from orca outputfiles that have a wrong syntax structure  when some numbers are to high.

PR #307, opened 2016-06-09
Add a --print option to the partial charge methods

ghutchis:
Prints atomic partial charges on standard output after calculating them


baoilleach:
Would you consider --printcharges instead to 'namespace' the option a bit better?

ghutchis:
Maybe. The trick is it's only with --partialcharges op, so --printcharges seemed a bit verbose. No big deal, though.

baoilleach:
Fair enough. I kind of wish though that we had a way of tying options for "ops" a bit closer to the particular "op". The current system feels just a bit too flexible. Maybe for OB 3000, we could have --pc:print etc. for all partial charge options.

PR #311, opened 2016-06-13
VASP Write. "w" and "z" options.

orex:
https://github.com/openbabel/openbabel/pull/188

### DIFF
--- a/src/formats/vaspformat.cpp
+++ b/src/formats/vaspformat.cpp
@@ -15,9 +15,10 @@ GNU General Public License for more details.
 #include <openbabel/babelconfig.h>
 #include <openbabel/obmolecformat.h>
 
-#include <limits.h>
+#include <limits>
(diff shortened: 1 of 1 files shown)


baoilleach:
Looks good. Thanks.

orex:
I also want to submit a test case for this, but the problem is, that I don't know the appropriate way to get a temporary folder files store within the test.

baoilleach:
If you can, it's best to avoid this problem by writing to a string instead of disk. You can just imagine people running the tests with "sudo" and creating all sorts of permissions problems. Could you describe the problem exactly - maybe I can help...?

orex:
The stream store is not working for VASP, because the information about structure spread over many files.

baoilleach:
Ah, I see. You may need to add a special option that writes everything to one file for testing purposes.

PR #312, opened 2016-06-17
Add deprecation warning for babel

baoilleach:
For "babel -H", the warning is added near the top of the help text
For "babel" just on its own, the warning is added at the end of the short usage information.
For other uses, no warning is given.

The text is:
<pre>
WARNING: babel is deprecated and will be removed in a future release
         of Open Babel. You should use obabel instead. For information
         on the differences please see:
             http://openbabel.org/docs/current/Command-line_tools
</pre>

### DIFF
--- a/tools/babel.cpp
+++ b/tools/babel.cpp
@@ -413,7 +413,7 @@ void deprecation()
   cout << "WARNING: babel is deprecated and will be removed in a future release" << endl
     << "         of Open Babel. You should use obabel instead. For information" << endl
     << "         on the differences please see:"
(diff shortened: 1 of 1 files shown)


baoilleach:
Drat - I got the URL wrong. I just had one thing to do, just one thing :-) Expect an updated pull request...

ghutchis:
Looks good. Anything else you want to get done before 2.4.0?

baoilleach:
I would like to sort out that api addition we discussed. I don't want to
delay things though.

PR #314, opened 2016-06-21
Remove OBAtom::IsThiocarboxylSulfur from the public API

baoilleach:
I've moved OBAtom::IsThiocarboxylSulfur to being a helper function in MOL2


baoilleach:
...I see a minor error. Expect an update.

PR #317, opened 2016-06-25
Fix comment symbol in swig interfaces

baoilleach:
Swig uses C style comments. Earlier I was using "#" sometimes which now causes a Swig error. Another dev had changed a few of these to %#. I have changed everything to //.


ghutchis:
I assume this is OK to merge for 2.4?

baoilleach:
Yes. It's just that recent swigs are more strict on syntax. Previously no errors or warnings, but now it refuses to build.

PR #319, opened 2016-06-28
Fixed a bug with using the --fillUC feature

psavery:
If pOptions is not NULL but no transforms are found, then
the SpaceGroup object (spacegroup) will contain nothing. This commit adds
a check so that pSG gets set to pUC->GetSpaceGroup() when pOptions is not
NULL but there are no transformations. Because of C++ short-circuits in
operators such as &&, we know that pOptions->find() will never be called
if pOptions is NULL.

### DIFF
--- a/src/ops/fillUC.cpp
+++ b/src/ops/fillUC.cpp
@@ -120,33 +120,31 @@ bool OpFillUC::Do(OBBase* pOb, const char* OptionText, OpMap* pOptions, OBConver
     obErrorLog.ThrowError(__FUNCTION__, "Cannot fill unit cell without a unit cell !" , obWarning);
     return false;
   }
-  
+
   OBUnitCell *pUC = (OBUnitCell*)pmol->GetData(OBGenericDataType::UnitCell);
   SpaceGroup spacegroup;
   const SpaceGroup* pSG;
   map<string,string>::const_iterator itr;
 
-  if(pOptions)
+  if(pOptions && pOptions->find("transformations") != pOptions->end())
   {
     itr = pOptions->find("transformations");
-    if(itr!=pOptions->end())
+    vector<string> vec;
+    tokenize(vec, itr->second.c_str());
+    for(vector<string>::iterator iter = vec.begin(); iter != vec.end(); ++ iter)
     {
-      vector<string> vec;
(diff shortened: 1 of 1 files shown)


psavery:
I added an extra condition to the if statement on line 129, but other than that, I mostly just removed the nested if statement inside it.

When using the cif file download from here: http://crystallography.net/cod/9009217.html

I wasn't able to get the "--fillUC" feature working until I updated the code like this. It looks like it isn't reading the commands - it may be a mmcif reader problem. Nevertheless, if there aren't any transformations found, we should still just use the transforms via GetSpaceGroup().

Another issue that may pop up in the future is that the atoms don't get sorted after filling the unit cell. As in, for MgSiO, after filling the unit cell, the atoms are arranged like MgSiOMgSiO - as in, the extra atoms don't get grouped with their original ones. This is probably not an issue for many formats, but it is an issue for some (like POSCAR).

ghutchis:
I think the atom grouping would need to be done by POSCAR or other output formats. I'm *not* going to reorder atoms in input or from --fillUC

ghutchis:
Otherwise, this looks fine.

psavery:
Yeah, that makes sense. I fixed the atom ordering issue for my POSCAR writer in Avogadro2 (it was an issue if a supercell was built, for example). I'll fix the one in OpenBabel sooner or later as well.

psavery:
Looks like someone already fixed it. But you have to add the "-xw" option or else it won't do it. Perhaps that should be default - I'm pretty sure the atoms are supposed to be grouped together for POSCAR format, although I could be wrong...

https://github.com/openbabel/openbabel/commit/2a2a97fa74f7f40453f4b3ae25ce396e3fc642e8

baoilleach:
Pinging @orex for comment on poscar.

orex:
Atoms sorting procedure is implemented in POSCAR plugin (write function). It should me no problem with the function from the proint of view. But IMHO it is better to collect symmetric sites in the same place as original one. It will be easy to edit check and edit output file manually.

One extra comment about the functionality. In disordered structures, where one atom can be substituted by another atom, two different type of atoms can occupy the same site. Does the code take care about it or not? I looked the code through, but did not find an answer.

PR #320, opened 2016-07-01
Removed 'SetSpaceGroup("P1")' from fillUC

psavery:
I'm not sure what this was here for, but it makes it so that when you
write to output files, the space group is "P1" instead of what it
actually is. Removing this line fixes that.


psavery:
That's strange... looks like it makes one of the tests seg fault.

Issue #325, opened 2016-07-18
Pybel not compatible with unicode_literals


mwojcikowski:
In Python 2.7 there is a bug when unicode_leterals are loaded

```
from __future__ import unicode_leterals
import pybel
m = pybel.readstring('smi', 'c1ccccc1')
```

```
TypeError                                 Traceback (most recent call last)
<ipython-input-3-c6b6881f6c6d> in <module>()
----> 1 m = pybel.readstring('smi', 'c1ccccc1')

/home/maciek/miniconda/lib/python2.7/site-packages/pybel.pyc in readstring(format, string, opt)
    195     obconversion = ob.OBConversion()
    196 
--> 197     formatok = obconversion.SetInFormat(format)
    198     if not formatok:
    199         raise ValueError("%s is not a recognised Open Babel format" % format)

TypeError: in method 'OBConversion_SetInFormat', argument 2 of type 'char const *'
```
Posible [solution](http://swig-user.narkive.com/pTuPgRC9/typemap-of-python-unicode-to-c-std-string-in-utf-8-encoding) (?):

baoilleach:
I don't really regard this as a bug. Support for unicode is not a feature.

mwojcikowski:
You can't write forward compatible code, which is a pain... Unless that's also a feature ;) Should we add this feature then? I think there is no downside to that, beside probably being a ns slower.

baoilleach:
If you want to. Are you saying that this also a problem in Python 3?

mwojcikowski:
No, Python 3.0 is fine. I think SWIG guess that in Python 2.x byte -> char and in Python 3.x unicode -> char.

mwojcikowski:
Fixed in #326

PR #326, opened 2016-07-18
[MRG] Support Python 2.x Unicode

mwojcikowski:
Fix for #325

### DIFF
--- a/scripts/openbabel-python.i
+++ b/scripts/openbabel-python.i
@@ -1,5 +1,9 @@
 %module openbabel
 
+%begin %{
+#define SWIG_PYTHON_2_UNICODE
+%}
+
 %{
 // used to set import/export for Cygwin DLLs
 #ifdef WIN32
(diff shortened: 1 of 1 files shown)


mwojcikowski:
The solution is much simpler than I've anticipated. There is `SWIG_PYTHON_2_UNICODE` variable that has to be defined to [support unicode in Python 2.x](http://www.swig.org/Doc3.0/Python.html). Although swig has to be at least 3.0.8 to work, which is the case for Anaconda - we should be fine. There shouldn't be any downside to this resolution for earlier versions.

mwojcikowski:
I've tested it on Anaconda build and it truly fix #325

baoilleach:
Can you confirm for sure that it doesn't cause a problem with earlier swigs?

mwojcikowski:
Travis use Swig 2.0.x ([log](https://travis-ci.org/openbabel/openbabel/jobs/145775600#L254)) and all tests pass. Defining arbitrary variable should not mess with anything (unless it is not meant for anything else). I may try building OB with Swig 3.0.7 which is just before that feature was used if you think it will help.

baoilleach:
Good point about Travis. Doh!

PR #329, opened 2016-07-21
Add support for 'M RGP' in mol files

baoilleach:
Numbered R groups (e.g. R1, R2) are currently handled like other atom aliases without special handling. However, to be interpreted as true R groups they need to be stored in an M RGP section. This code allows these sections to be read and written.

Note to self: I should document how to set these through the API.

(I intend to write a test case or two for this code, so maybe don't merge until then unless the release is imminent)


ghutchis:
I'm not going to merge until you write the test case. Either way, I'm cherry-picking fixes for 2.4, so I'll merge when you're ready.

baoilleach:
That's the test cases done unless you can see something I'm missing.

For the most part the new code is just special-casing the instance where an atom alias is Rn, where n is an integer. The existing code had several ways of entering such aliases - hence the multiple test cases.

PR #330, opened 2016-07-21
Fixed bug related to charges from cdx file

genevievebo:
Problem reading the charges from the CDX file (e.g. M  CHG  2   432513   832767  instead of M  CHG  2   4   1   8  -1;  value of 32767 instead of -1).

### DIFF
--- a/src/formats/chemdrawcdx.cpp
+++ b/src/formats/chemdrawcdx.cpp
@@ -686,7 +686,7 @@ puts("found name");
 
   int getCharge(istream *ifs, UINT32 size)
   {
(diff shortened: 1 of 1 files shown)


baoilleach:
"int charge = 0;" perhaps? Is there some reason the BIG_ENDIAN test was removed?

baoilleach:
This code doesn't compile on our test systems. See under "All checks have failed".

ghutchis:
@baoilleach - it looks like the chemdraw code was rewritten - is this still needed?

PR #331, opened 2016-08-02
restore NumRotors behavior and add rotor count output to obprop

dkoes:
For assessing ligand flexibility, counting rotors within rings is highly undesirable.  This patch restores the old behavior of NumRotors and makes the new behavior available with a boolean argument.

Internally, there don't appear to be any dependences in NumRotors, so restoring the old behavior will probably keep the most people happy.

### DIFF
--- a/include/openbabel/mol.h
+++ b/include/openbabel/mol.h
@@ -279,8 +279,8 @@ enum HydrogenType { AllHydrogen, PolarHydrogen, NonPolarHydrogen };
     unsigned int NumHvyAtoms();
     //! \return the number of residues (i.e. OBResidue substituents)
     unsigned int NumResidues() const      { return(static_cast<unsigned int> (_residue.size())); }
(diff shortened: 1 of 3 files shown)


baoilleach:
That was a quick merge. I was going to suggest that pushing this change into OBBond::IsRotor() instead would be a better fix as we need this option there also.

dkoes:
So move the sampleRingBonds check in FindRotors inside of IsRotor and pass it sampleRingsBonds in FindRotors instead?

That is, instead of:
```
    if (bond->IsRotor()) {
       ...
        if (bond->IsInRing()) {
          if (!sampleRingBonds)
            continue; // ignore it
```
do
```
    if (bond->IsRotor(sampleRingBonds)) {
       ...

```
And then have NumRotors do what it did before and call IsRotor.

dkoes:
OBRotorList::SetEvalAtoms also calls IsRotor, but does not have a sampleRingBonds bool - what is the right default for this function?

PR #332, opened 2016-08-03
Bugfix for multi-model PDB writing

Jsunseri:
Bugfix for multi-model PDB writing. Previous versions inserted an
'END' tag before the 'ENDMDL' tag after each model. This violates
the PDB spec, which says 'END' marks the end of file, and also
trips up some programs/APIs (e.g. rdkit) that follow the spec and
stop reading the resulting PDB files after the first model.
Minimally invasive fix. I can make a test if desired.


ghutchis:
I'm happy to merge - this looks fine, but a regression test would be ideal. Thanks!

PR #333, opened 2016-08-04
enable sampleRingBonds option in IsRotor as well

dkoes:
Code modifications in response to @baoilleach comments on #331 for your consideration.


baoilleach:
In IsRotor, can you rename the param from "sampleRingBonds" to
"includeRingBonds" (or indeed "excludeRingBonds") or so? The term sampling
is ambiguous in this context.

PR #339, opened 2016-09-12
Added line for -NO2 group

ntropia2:
Added an entry for the -NO2 group to preserve the explicit charge localization [N+](=O)[O-] when adding hydrogens at specific pH.


data/phmodel.txt
```diff
@@ -47,6 +47,7 @@ TRANSFORM [NH1:1]~[ND2:2]~[NH1:3]~* >> [N-:1]=[N+:2]=[N:3]-* 1E+10	# azide group
 
 #nitro group (-NO2)
 TRANSFORM [O:1]=[N:2]-[O:3] >> [O:1]=[N+:2]-[O-:3]  1E+10
+TRANSFORM [O:1]=[N:2]=[O:3] >> [O:1]=[N+:2]-[O-:3]  1E+10
```
ghutchis: This one won't fly. If people have explicit hypervalent N, I don't want to suddenly force their bonding to match a different representation.

ghutchis:
I'll reiterate. Can you supply me with a version that does not change the explicit hypervalent O=N=O version? If someone has picked that representation, I'm not going to change it on them.

PR #343, opened 2016-09-15
add support for customizations of filter and rotors used during conformer search and small bugfixes

razziel89:
When performing conformer searches on large molecules, it might be beneficial to a) fix certain rotors so that they are not changed and b) make the steric overlap filter less stringent. I made both changes available to the command-line interface.

I also noticed that the help message said a switch were called "--converge" but the code actually expects "--convergence". It also sometimes happens that none of the initial conformers pass the filter and the programme would then run indefinitely. Those bugs are also fixed.

### DIFF
--- a/include/openbabel/conformersearch.h
+++ b/include/openbabel/conformersearch.h
@@ -3,6 +3,7 @@ conformersearch.h - Conformer searching using genetic algorithm.
 
 Copyright (C) 2010 Tim Vandermeersch
 Some portions Copyright (C) 2012 Materials Design, Inc.
+Some portions Copyright (C) 2016 Torsten Sachse
 
 This file is part of the Open Babel project.
 For more information, see <http://openbabel.org/>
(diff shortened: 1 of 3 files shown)


ghutchis:
I'm fine to merge this, but the AppVeyor (Visual C++) build must be fixed:

```
C:\openbabel\src\ops\conformer.cpp(98): error C2440: 'initializing' : cannot convert from 'std::basic_istream<char,std::char_traits<char>>' to 'bool' [C:\openbabel\build\src\plugin_ops.vcxproj]
          No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
          C:\openbabel\src\ops\conformer.cpp(125) : see reference to function template instantiation 'bool OpenBabel::getValue<int>(const std::string &,T &)' being compiled
```

razziel89:
Thanks! This most recent commit should have fixed the build error. Shall I close this now or after you performed the merge?

ghutchis:
Thanks!

PR #344, opened 2016-09-15
track the number of electrons cored by an electron core potential and read that information from Molden-files

razziel89:
Added methods to get and set the number of electrons cored by an electron core potential.

Molden files store an element via both the element symbol and the element number. If the element number is replaced by the actual charge (that not cored by an ECP), the number of cored electrons can be inferred directly from the file.

### DIFF
--- a/include/openbabel/atom.h
+++ b/include/openbabel/atom.h
@@ -101,6 +101,7 @@ namespace OpenBabel
       OBResidue                    *_residue;   //!< parent residue (if applicable)
 
       unsigned long                 _id;        //!< unique id
+      unsigned int                  _ecp;       //!< number of electrons cored by ecp
 
       //! \return All flags
       int  GetFlag() const    {  return(_flags);  }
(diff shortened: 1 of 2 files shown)


ghutchis:
IHMO, this doesn't make much sense in Open Babel. I think it's a better fit for [cclib](http://github.com/cclib/cclib/) 

I'm fine with adding atom properties, but since this is going to be a specialized case, you're better off using OBPairData or something similar to store the core ECP into the OBAtom class. I'm not going to merge a patch that adds storage for every OBAtom in a PDB file - it just doesn't make much sense.

razziel89:
You are correct, it's probably not worth the added storage in general. Thanks for pointing out OBPairData, I hadn't known of it until now. I will have a look at how to use it for this. Would you recommend filling these data in moldenformat.cpp or completely separately (e.g. via language bindings)?

I am still wondering about what should happen in cases where the element symbol and the atomic number don't agree. I have a couple of molden files where this is the case because the atomic number is reduced by the number of cored electrons. So far, Open Babel uses the atomic number and disregards the element symbol, which caused converted geometries to have mostly wrong elements. I am unsure as to whether the element symbol should take precedence (this would be my favourite version) or whether an exception or warning should be thrown. What do you think?

ghutchis:
I'd use the element symbol if it will be more reliable. It's probably also worth issuing a warning. I'll leave it up to you - if you want to close the patch and give a new one, that's fine. Or you can fix it and use OBPairData in the OBAtom objects and submit here.

razziel89:
I decided to use the element symbol and added a warning. Do you think this can be merged?

PR #350, opened 2016-10-04
Add support for further manipulations to OBMol 

razziel89:
These modifications include:
- angles
- dihedral angles (please see comment below)
- mirroring + inversion
- translation (also of subunits)
- rotation (also of subunits)
- aligning a molecule to a plane in space
- cleaving part of a molecule
- bonding two molecules together

The method "SetTorsion" does the same as the newly introduced "SetDihedralAngle". However, I prefer the interface of "SetDihedralAngle" since the atoms are specified in terms of indices and the angle is specified in degrees. Hence, I would like to suggest adding a method with such an interface.

I would also like to submit my code for the treatment of aggregates (i.e., manipulating single molecules within an aggregate) and my code to screen molecules by geometrical similarity However, since these bits of code rely on the modifications introduced here, I wanted to submit them first.


baoilleach:
May I suggest that instead of adding new methods to OBMol, you create a new class to handle such manipulations, e.g.

manipulator = OBMoleculeManipulator(myOBMol)
manipulator.InvertGeometry()

razziel89:
This sounds like a really good suggestion for cleaving and bonding. However, I would not say it's a good way to go for the rest of the manipulations. Firstly, this would separate the methods to get and set certain properties as they would reside in different classes. I think this would make using them more difficult and intransparent. Secondly, if only the methods introduced by my pull request were to be moved to a new class, this would leave some manipulation methods in OBMol, making the separation rather arbitrary, IMHO. It seems as if moving all set methods as well as the corresponding get methods (and the bonding and cleaving methods) to a new class might be the best choice. Do you agree?

baoilleach:
We cannot remove any methods from the current API, no matter how much we want to. So, sure, I agree with the second option, but we can't do that. Similarly, we can't replace SetTorsion with SetDihedralAngle.

razziel89:
OK, so that is not an option then. The only remaining viable option that I can see is to create a new manipulator class that contains all the functionality (in a consistent way) whilst leaving the current API unchanged. Access to the functionality already part of OBMol could be realized via wrapper methods in the new class. That way, the current API remains fully intact while avoiding an arbitrary separation of functionality. One might add a "warning" (possibly in comment-form) to the manipulation methods that are currently part of OBMol to tell the users not to use them anymore if it can be avoided. Obviously, whether or not to introduce such a warning is not for me to decide.

Also, I did not mean that SetDihedralAngle should replace SetTorsion, but rather that it be provided as an alternative to SetTorsion as the latter is a bit awkward to use via the language bindings, IMHO. It could be provided within the new class if that route is taken.

ghutchis:
I think a new manipulator class would be great. As Noel mentioned, we've said "no API breaks" so some methods will remain, albeit deprecated.

razziel89:
OK, sounds good. I just don't know when I'll find the time to actually implement that since there are some important deadlines ahead, so it might take some weeks. However, I have still one question left: What do you consider an API-break? Does adding a new but optional parameter to an existing function count as one?

baoilleach:
FYI, the rule on API breakage has been relaxed for the next release, with the idea of tidying up the current API and fixing long standing issues.

razziel89:
Great, thanks for the information. Do you already know when the next release is due to be published? I'll try to make all the necessary changes before then.

razziel89:
I outsourced all new and some old manipulation routines to a new class `OBMolManip`. If this can be merged, I would be happy to provide more new functionality for, e.g., treating aggregates, screening conformers by geometrical similarity, parallel determination of pointgroups, etc, all outsourced properly. However most of that code relies on some of the here-introduced manipulations.

ghutchis:
That sounds cool. I'll try to review this soon.

As far as point groups, I'm going to migrate OB to use the new libmsym, which I've found to be faster and more feature-rich. This work will happen sometime in May/June.
https://github.com/mcodev31/libmsym

anon3742683721 (changes requested):
visual lewis dot plugin; auto rotation visual; dimensional analysis

razziel89:
Closing this PR as I have moved away from projects involving `openbabel`. It's a great project and I wish it al the best. It just does not fit my use cases anymore.

PR #357, opened 2016-10-27
Fix for missing ZLIB on win32

philthiel:
I'm not entirely sure, but I suppose this check for ZLIB availability is necessary.
Otherwise, CMake throws an error due to missing variable 'ZLIB_LIBRARY'.


ghutchis:
Well, it look fine and passes the AppVeyor Windows build, so I'm willing to merge. Thanks!

PR #360, opened 2016-12-03
Improve svg ball and stick

ghutchis:
Before:
screen shot 2016-12-02 at 11 55 12 pm

After:
screen shot 2016-12-02 at 11 58 19 pm


ghutchis:
Another example - before:
screen shot 2016-12-03 at 12 02 36 am

After:
screen shot 2016-12-03 at 12 03 10 am

baoilleach:
Actually - I quite like the before also. Would be great to have an option to switch between.

PR #361, opened 2016-12-11
Avoid IndexError for plugins with empty names

langner:


mbanck:
Hi, I got an off-list request that this is also broken in 2.4 so it seems to affect users quite heavily. Should this be backpated to the openbabel-2-4-x branch?

ghutchis:
Yes - I’ve been swamped but should be able to do that today.

PR #364, opened 2016-12-14
Add new phmodel rules

drc007:
Added new rules for other functional groups.


baoilleach:
I'll get to this soon. But do you want to tighten up your pyridine pattern a bit? Currently it hits molecules like the following - is this as you expected?:
pyridine

drc007:
Many of those structures are not aromatic rings? Should they be hit by a SMARTS that only contains aromatic atoms?

Chris

baoilleach:
They are aromatic according to the Daylight aromaticity model. I was thinking more about whether you want to exclude those cases where there is a carbonyl adjacent to the nitrogen. I don't know - maybe not. Might see you on Wednesday?

drc007:
I plan to be there on Wed,

Chris

PR #367, opened 2017-01-08
Fixed a few errors in space-groups.txt

psavery:
As was pointed out in issue number 153 on
https://github.com/openchemistry/avogadrolibs, there
are a few issues in the space groups. There were a few
illegal '#' symbols and potential problems with Hall numbers
506 (international number 210) and 450 (international number
160). Unfortunately, I couldn't fix 450, but I was able to
fix 506. A test was performed: after using the transforms
in Avogadro, spglib identified the space group correctly.
For some reason, this was not the case for number 450...


psavery:
Unfortunately, as I mention in the commit, Hall Number 450 might still be broken. This would potentially be correct for it: http://www.cryst.ehu.es/cgi-bin/cryst/programs/nph-getgen?what=text&gnum=160

However, it isn't identified as the correct space group by spglib when I use it. For number 506, though, I was able to fix it and it is identified as the correct space group.

ghutchis:
I see that there are some updates to spglib (currently 1.9.9). Maybe it's worth discussing with them? Certainly we all want the correct matrix / Hall number assignments.

ghutchis:
I merged for now and then let's revisit HN 450 with spglib.

psavery:
Well, I don't think spglib has the transforms. I don't know completely how it works, but I know it detects symmetry in the crystal and then assigns a space group based off the symmetry. The transforms are usually used for generating a crystal with symmetry (like filling a unit cell).

I was using Bilbao to find the transforms. For example, for space group 210, here is one: http://www.cryst.ehu.es/cgi-bin/cryst/programs/nph-getgen?what=text&gnum=210

However, it is somewhat tricky to do. Bilbao only has the space groups organized as the 230 fundamental space group numbers. In Open Babel and spglib, though, we have the space groups organized as 530 hall numbers (several of which are just a different representation of the same space group, which is why there are extra). In some cases, there is only one Hall number that matches an international number, so that is easy to pair. other cases may not be, though.

For the tests I did, I was filling a unit cell with Avogadro using the Open Babel transforms, and then I was just detecting the space group with spglib to make sure the right space group was being detected.

Anyways, I think it'd be better if we found somewhere else to get the transforms from so it'd be less complicated. I just don't know where yet...

PR #370, opened 2017-01-30
Include original when there are zero rotatable bonds in confab

cowsandmilk:
If you don't include the original coordinates, confab currently will give you no conformers for molecule with zero rotatable bonds, so the molecule disappears from the output file. This makes confab return the original coordinates if there are zero rotatable bonds.


baoilleach:
Probably a good idea. :-) If could rewrite to use a straightforward "if" statement I think the intention would be clearer and avoid the self-assignments.

Issue #368, opened 2017-01-09
Python3.6 openbabel: No module named 'DLFCN'


klniu:
In the new release Python 3.6, the DLFCN is removed, see https://docs.python.org/3.6/whatsnew/3.6.html

```
File "/usr/lib/python3.6/site-packages/openbabel.py", line 6, in <module>
    import DLFCN as dl
ModuleNotFoundError: No module named 'DLFCN'
```

mwojcikowski:
Two questions:

1. Is it cross-platform compatible?
2. Is it backward compatible?

baoilleach:
I would be open to just putting the constant in there directly. After all, there's no need to import the whole of ctypes just for the number '4' or whatever.

cowsandmilk:
Isn't the point that the value is different on different platforms?

For instance, on my mac, it is 8. On my ubuntu system, it is 256.

baoilleach:
That would be a problem alright. Ignore my prevoius comment so.

Ultimately the issue is that on some platforms, the DL loader fails to load the plugins without this kludge. This is also the case for other bindings and I think the correct solution is to move towards statically including the plugins in the main .so, so to avoid this issue completely.

mwojcikowski:
Also the problem with `RTLD_GLOBAL` is that it messes up some other Python libraries which do not anticipate `RTLD_GLOBAL`. Prime example is L-BFGS-B from scipy which segfaults when loaded after OB.

baoilleach:
That's right. But it's only needed for plug-in load, and can be turned off afterwards. Maybe the bindings should instantiate an OBConversion object to force plugin loading, and then turn it off.

mwojcikowski:
I can confirm that calling following code after importing openbabel and/or pybel "unbreaks" other C libraries:

> ob.OBConversion()
> sys.setdlopenflags(sys.getdlopenflags() ^ RTLD_GLOBAL)

Pybel does initialize OBConversion(), I repeated it just to be sure (not all may want to use friendly interface).

hseara:
Shall we close this bug after the merge? If further cleaning wants to be done, it can be done aside. It is pleasant to keep the bug/tracker clean.

baoilleach:
Yes - this should have been closed after the merge.

PR #372, opened 2017-02-02
Fix bug #368: Python3.6 openbabel: No module named 'DLFCN'

hseara:
The DLFCN module has been removed from python 3.6, as it is not
documented. Same funtionality can be achive with the os module
that makes available the os.RTLD_GLOBAL variable for dlopen()

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -81,11 +81,8 @@ if (DO_PYTHON_BINDINGS)
             COMMAND ${SWIG_EXECUTABLE} -python -c++ -small -O -templatereduce -naturalvar -I${openbabel_SOURCE_DIR}/include -I${openbabel_BINARY_DIR}/include -o ${openbabel_SOURCE_DIR}/scripts/python/openbabel-python.cpp ${eigen_define} -outdir ${openbabel_SOURCE_DIR}/scripts/python ${openbabel_SOURCE_DIR}/scripts/openbabel-python.i
               COMMAND ${CMAKE_COMMAND} -E echo "import sys" > ob.py
               COMMAND ${CMAKE_COMMAND} -E echo "if sys.platform.find('linux'\) != -1:" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "  try:" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "    import dl" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "  except ImportError:" >> ob.py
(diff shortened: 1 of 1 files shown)


scripts/CMakeLists.txt
```diff
@@ -81,11 +81,8 @@ if (DO_PYTHON_BINDINGS)
             COMMAND ${SWIG_EXECUTABLE} -python -c++ -small -O -templatereduce -naturalvar -I${openbabel_SOURCE_DIR}/include -I${openbabel_BINARY_DIR}/include -o ${openbabel_SOURCE_DIR}/scripts/python/openbabel-python.cpp ${eigen_define} -outdir ${openbabel_SOURCE_DIR}/scripts/python ${openbabel_SOURCE_DIR}/scripts/openbabel-python.i
               COMMAND ${CMAKE_COMMAND} -E echo "import sys" > ob.py
               COMMAND ${CMAKE_COMMAND} -E echo "if sys.platform.find('linux'\) != -1:" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "  try:" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "    import dl" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "  except ImportError:" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "    import DLFCN as dl" >> ob.py
-              COMMAND ${CMAKE_COMMAND} -E echo "  sys.setdlopenflags(sys.getdlopenflags() | dl.RTLD_GLOBAL)" >> ob.py
+              COMMAND ${CMAKE_COMMAND} -E echo "  import os" >> ob.py
+              COMMAND ${CMAKE_COMMAND} -E echo "  sys.setdlopenflags(sys.getdlopenflags() | os.RTLD_GLOBAL)" >> ob.py
```
ghutchis: I don't believe this will work on Python 2.x. It looks like it goes back to Python 3.3, though.
https://docs.python.org/2.7/library/os.html#module-os

ghutchis:
This looks ok to me @baoilleach - does it look ok to you?

ghutchis:
Bump @baoilleach - is this an OK fix for Python 3.6?

hseara:
I see no re problem with my fix. We are using it on all our computers and so far no problems. This fix works in Python 2.7 and Python 3.6, and I suggest we merge it. If further work to get rid of this code wants to be done later on it is fine for me. Until then is a real conservative fix.

baoilleach:
Sorry. Will review...

baoilleach:
Apologies for the delay - it slipped through my indox. Merging.

PR #377, opened 2017-02-07
OBRandom::NextGaussian()

arkose:
I needed normal distribution random numbers for a personal project, so I added a NextGaussian() method to the OBRandom class. I think it may be useful to others.

### DIFF
--- a/include/openbabel/rand.h
+++ b/include/openbabel/rand.h
@@ -90,6 +90,9 @@ namespace OpenBabel
     int NextInt();
     //! \return a random floating-point number between 0.0 and 1.0
     double NextFloat();
+    //! \return a random Gaussian value with mean @p mean
+    //! and standard deviation @p std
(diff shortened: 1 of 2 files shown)


baoilleach:
IMO, a random number generator has no place in a cheminformatics library, and we should remove it rather than add to it. The existing code is a pseudo-random number generator with a low repeat, but presented as an actual random number generator.

ghutchis:
I'd agree 100% with Noel. I'd suggest some of the new C++11 features, e.g.
http://stackoverflow.com/a/7114482/131896

```
std::uniform_int_distribution<uint32_t> uint_dist;         // by default range [0, MAX]
std::uniform_int_distribution<uint32_t> uint_dist10(0,10); // range [0,10]
std::normal_distribution<double> normal_dist(mean, stddeviation);  // N(mean, stddeviation)
```

PR #381, opened 2017-02-14
Improve thread safety for global objects

baoilleach:
OBAromaticTyper, OBPhModel, OBAtomTyper, OBRingTyper are now all thread-safe (I believe), with only minimal changes needed for all but OBAromaticTyper.

I've also tided up the use of some externs in residue.cpp/.h, which were probably not a problem but it's good to eliminate suspects from our enquiries.

NOT DONE: chains.cpp has several static variables, which will cause problems. This is a bit of work; hence, this pull request with the low-hanging fruit.


baoilleach:
In future I'll merge my own pull requests after, let's say, 7 days. I'll request a review if I really feel someone ought to check it out.

PR #386, opened 2017-02-22
Add an option to the canonical SMILES format to specify the timeout.

timvdm:


baoilleach:
It's good to merge, but it's still worth noting that it shouldn't take 5 seconds to canonicalise this structure, let alone 60s. I don't know if Tim is in a position to improve the algorithm at the moment, but John here in the office has done some cool stuff in this area, and the day might come when we follow that up.

timvdm:
This should be possible and I will have a look at this.

A quick look at the code shows that both the algorithm (use orbits of the permutation group) and performance can be improved (e.g. using a small vector optimization for inner loops, avoid dynamic allocated memory in general).

timvdm:
Hi,

I did have a look at this but didn't get far. The OB algorithm is still
based on a depth-first search while the better algorithms are based on
partition refinement. This allows the automorphisms to be used much more
effectively. Such an algorithm would be better for OB but this would change
the canonical order again (i.e. not backwards compatible). It could be
added as a can2 format though.

Kinds regards,
Tim

baoilleach:
I'd be okay with changing the canonical order if it's an improvement.... Also, I might be able to convince John here to make available some of his code which is better than partition refinement. Let me have a chat with him...

baoilleach:
Just remembered that I left this comment hanging. To finish my thoughts:

Already as part of the kekulization changes the SMILES output has changed. For example, I write out a bond symbol between aromatic atoms if the bond in is a ring and is not itself aromatic. So don't let changing the canonical order hold you back - improving the speed of the canonicalisation would be great. There are some other features of the current approach that would be great to change if you decide to do this (interested in writing an OB Enhancement Proposal?).

Regarding John's canonicalisation, he hopes to write this up but nothing is going to happen in the short term - he's busy working on some CIP stuff.

PR #1544, opened 2017-02-24
Added OBAtom::GetAtomID() method.

arkose:
Discussion on the 'openbabel-discuss' mailing list asked how to get the text id
of an OBAtom; the current approach requires getting the atom's residue to call
OBResidue::GetAtomID(OBAtom*) and passing in the atom's pointer. To simplify
this use case, OBAtom::GetAtomID() is added by this patch.


baoilleach:
This looks like a convenience function. We should be removing these from classes not adding them. Let's discuss on the list a better place for all convenience functions.

PR #1546, opened 2017-02-26
Fastsearch 64 fix

dkoes:
The fastsearch index was using 32-bit file offsets, which meant it couldn't be used on input files with size larger than 4GB.  This switches to using 64-bit file offsets.  (Fastsearch indices don't store the actual molecular data, instead storing a file offset into the original file for openbabel to re-parse from). Backwards compatibility is achieved by repurposing one byte of the fpid field (which will always be zero on older versions).  I don't think any fingerprints have names longer than 13 bytes, so I don't think this is a problem.

There is a still a 2^32 limit on the number of molecules.  Removing this limit would require a more invasive change to the header format.

I've checked (briefly) that this still works with old databases while successfully searching all of pubchem (94,453,395 molecules in 2.5 minutes) when the index is built with the new code.


ghutchis:
I'd be open (for v3) to go beyond 2^32 in the header. I'll merge this for now - looks good.

PR #1547, opened 2017-03-11
Don't try to install aromatic.txt as it is no longer present

baoilleach:
A fix for my merge of pull request #1545  which caused "make install" to fail as aromatic.txt was no longer present


ghutchis:
Agreed.

Issue #1552, opened 2017-03-24
Python SWIG binding OBResidue::GetAtoms() broken


djhogan:
This may be related to issue #1551, but I tried using OBResidue's GetAtoms method because I'm trying to get the character code of the atoms in each residue. I can call the method and assign the result to a variable, but the result can't be indexed (has no __getitem__ attribute). Also, when I evaluate the variable in the interpreter, I get a message about a memory leak. I've tried demonstrating this behaviour below.

```
>>> r = mol.GetResidue(2)
>>> r.GetNumAtoms()
16
>>> r.GetAtoms()
<Swig Object of type 'std::vector< OpenBabel::OBAtom *,std::allocator< OpenBabel::OBAtom * > > *' at 0x105e14450>
>>> atom_list = r.GetAtoms()
>>> atom_list
swig/python detected a memory leak of type 'std::vector< OpenBabel::OBAtom *,std::allocator< OpenBabel::OBAtom * > > *', no destructor found.
<Swig Object of type 'std::vector< OpenBabel::OBAtom *,std::allocator< OpenBabel::OBAtom * > > *' at 0x105e14420>
>>> atom_list[1]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'SwigPyObject' object has no attribute '__getitem__'
```

baoilleach:
Again, that method should be commented out in the bindings, as the atoms should be accessed instead through an interator, OBResidueAtomIter.

baoilleach:
Fixed by #1554.

PR #1553, opened 2017-03-25
Added std::vector<OBAtom> and std::vector<OBAtom *> bindings to python SWIG.

djhogan:
To fix issue #1552. I added another operator= method to OBAtom that accepts a const argument because the original operator= method accepts a non-const reference to OBAtom and this broke vector templates that needed a const version. So, I added the const version, keeping the original version for backwards compatibility.


baoilleach:
There's no need for these changes as discussed over at #1552.

Issue #1551, opened 2017-03-24
No python SWIG binding for std::vector of OBAtom *


djh901:
Would it be as easy as adding a line to this part of openbabel-python.i?

```
VECTORTEMPLATE_WRAP(Int, int)
VECTORTEMPLATE_WRAP(UnsignedInt, unsigned int)
VVTEMPLATE_WRAP(Int, int)
VECTORTEMPLATE_WRAP(Double, double)
VECTORTEMPLATE_WRAP(ULong, unsigned long)
VECTORTEMPLATE_WRAP(String, std::string)
VECTORTEMPLATE_WRAP(Vector3, OpenBabel::vector3)
VVTEMPLATE_WRAP(Vector3, OpenBabel::vector3)
VECTORTEMPLATE_WRAP(OBMol, OpenBabel::OBMol)
VECTORTEMPLATE_WRAP(OBBond, OpenBabel::OBBond)
VECTORTEMPLATE_WRAP(OBResidue, OpenBabel::OBResidue)
VECTORTEMPLATE_WRAP(OBRing, OpenBabel::OBRing)
VECTORTEMPLATE_WRAP(pOBRing, OpenBabel::OBRing*)
VECTORTEMPLATE_WRAP(pOBGenericData, OpenBabel::OBGenericData*)
VECTORTEMPLATE_WRAP(pOBInternalCoord, OpenBabel::OBInternalCoord*)
```

ghutchis:
Yes, that's right.

baoilleach:
What function are you calling?

djh901:
FindChildren. It seems to be overloaded with one version supporting an integer vector and the other supporting a vector of atom pointers. I wanted to use the latter version, but found there was no way to create a vector of Atoms. I did some digging to figure out why and found the above code that curiously omits OBAtom and OBAtom * from std::vector template bindings.

baoilleach:
Every addition to std::vector in the bindings creates a lot of boilerplate
code and so I've avoided needlessly adding methods where a duplicate
exists. From what you say, this is also the case here. From a user
perspective, the other method should be excluded from the bindings
explicitly to avoid confusion.

djh901:
Should I add an %ignore line to the interface files, then?

baoilleach:
That'd be great. Is it okay to close this bug report so...?

PR #1554, opened 2017-03-26
Ignore some non-functioning python SWIG bindings.

djhogan:
Refer to issues #1551, #1552, and closed pull request #1553.


baoilleach:
Looks good. Thanks for the patch.

Issue #1555, opened 2017-03-27
Heap corruption if *.nwo.file is converted twice with obgui.exe


yodah134:
- [x ] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 2.4.1 (32 and 64 bit)
Operating system and version: Windows 10 64-bit

## Expected Behavior
If a NWChem output file is converted with obgui.exe to any other file type (e.g. *.xyz) this works fine when the 'CONVERT' button is clicked the first time.

## Actual Behavior
If the button is clicked again obgui.exe crashes with a heap corruption exception (error code 0x000374).
The problem is present in the Windows versions of obgui.exe (32 bit and 64 bit) delivered with OpenBabel 
2.4.1
The problem does not occur if obabel.exe is called twice.
The problem was not present in obgui.exe delivered with OpenBabel 2.3.2.
This is not trivial: the same error occurs if a command of type 'OBConv.ReadFile(OBMol, MolFilePath)' is called twice in an application if accessed e.g. via ObDotNet.dll thus crashing the calling process.

## Steps to Reproduce
[Acetonitrile.txt](https://github.com/openbabel/openbabel/files/872762/Acetonitrile.txt)
(change file extension from txt to nwo)

djhogan:
I reproduced this on my mac (OpenBabel 2.4.1, OSX 10.12.1). It crashes when I try to convert Acetonitrile.nwo to *.xyz on the first try.

baoilleach:
If you want to try to fix it, create a debug build, and then run it through
valgrind. It's often fairly obvious what the problem is when you see the
line where it occurs.

PR #1556, opened 2017-03-29
Remove delete statement.

djhogan:
This may or may not fix issue #1555, but I noticed it results in an unallocated pointer being freed, crashing the program.

Details:
When from_scratch is true, coordinates is not allocated. A separate if
statement was added to handle the case when from_scratch is true that
does not try to free coordinates.


baoilleach:
Looks good to merge, but would be good to confirm situation re #1555. Is
this possible?

djhogan:
obgui doesn't crash for me anymore when I try to reproduce the error in #1555. It doesn't crash my test program anymore either

Issue #340, opened 2016-09-13
Total spins in -oreport are wrong in many systems


nakatamaho:
- K do not output total spin and it should. It is a doublet system
  $ obabel -:"[K]" -oreport
  FILENAME:
  FORMULA: K
  MASS: 39.0983
  EXACT MASS: 38.9637067
  INTERATOMIC DISTANCES
- Following AlO should have TOTAL SPIN 2 (or maybe 4)
  $ obabel -:"O=[Al]" -oreport
  FILENAME:
  FORMULA: AlO
  MASS: 42.9809
  EXACT MASS: 42.9764532
  TOTAL SPIN: 3
- Following [PH2-] system should have TOTAL SPIN 1 or 3
  obabel -:"[PH2-]" -oreport
  FILENAME:
  FORMULA: H2P-
  MASS: 32.9896
  EXACT MASS: 32.9894117
  TOTAL CHARGE: -1
  TOTAL SPIN: 2
- Na atom should be doublet
  $ obabel -:"[Na]" -oreport
  FILENAME:
  FORMULA: Na
  MASS: 22.9898
  EXACT MASS: 22.9897693
- Li atom should be doublet
  obabel -:"[Li]" -oreport
- C2H5AlCl+ should be 
  $ obabel -:"CC[Al+]Cl" -oreport
  there are 22 electrons, but reports TOTAL SPIN: 2
  FILENAME:
  FORMULA: Li
  MASS: 6.9410
  EXACT MASS: 7.0160045
  INTERATOMIC DISTANCES
- LiH- should be doublet
  $ obabel -:"[H-].[Li]" -oreport
  FILENAME:
  FORMULA: HLi-
  MASS: 7.9489
  EXACT MASS: 8.0238296
  TOTAL CHARGE: -1

INTERATOMIC DISTANCES
INTERATOMIC DISTANCES

baoilleach:
I want to add a reference here to https://github.com/openbabel/enhancement-proposals/pull/3. The code for atom-based spin multiplicity is undergoing some revisions at the moment.

PR #1562, opened 2017-04-29
Total spin should be calculated by formal charge.

nakatamaho:
This fixes following issue.
https://github.com/openbabel/openbabel/issues/340
The total spin of the ground state is difficult to calculate, however, my simple patch
is better than apparently wrong total spin.

### DIFF
--- a/src/formats/qchemformat.cpp
+++ b/src/formats/qchemformat.cpp
@@ -345,8 +345,12 @@ namespace OpenBabel
 
     FOR_ATOMS_OF_MOL(atom, mol)
       {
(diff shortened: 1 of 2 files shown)


nakatamaho:
I also added a patch for better qchem format. it reads "%3d%15.5f%15.5f%15.5f" 
cf. https://www.q-chem.com/qchem-website/manual/qchem43_manual/sect0014.html

baoilleach:
Thanks. We would appreciate separate pull requests for independent changes. This makes it easier reviewing, writing changelogs, etc.

PR #1563, opened 2017-04-29
better calculation for total spin

nakatamaho:
This fixes following issue.
#340
The total spin of the ground state is difficult to calculate, however, my simple patch
is better than apparently wrong total spin.

### DIFF
--- a/src/mol.cpp
+++ b/src/mol.cpp
@@ -1176,29 +1176,26 @@ namespace OpenBabel
   }
 
   //! Returns the total spin multiplicity -- if it has not previously been set
-  //!  It is calculated from the atomic spin multiplicity information
-  //!  assuming the high-spin case (i.e. it simply sums the number of unpaired
-  //!  electrons assuming no further pairing of spins.
+  //! calculating from atomic formal charge
   unsigned int OBMol::GetTotalSpinMultiplicity()
   {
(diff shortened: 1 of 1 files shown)


ghutchis:
I'm OK with this style of calculation if no atomic spins are set. However, the previous calculation is useful, e.g., radicals from SMILES or MDL Molfiles, etc. If I have set different atoms as having radical (atomic spin) character, I want to preserve that in the molecular total spin calculation.

src/mol.cpp
```diff
@@ -1176,29 +1176,26 @@ namespace OpenBabel
...
         OBAtom *atom;
         vector<OBAtom*>::iterator i;
-        unsigned int unpaired_electrons = 0;
+        int chg = 0;
+        int spin = 0;
 
         for (atom = BeginAtom(i);atom;atom = NextAtom(i))
-          {
-            if (atom->GetSpinMultiplicity() > 1)
-              unpaired_electrons += (atom->GetSpinMultiplicity() - 1);
```
ghutchis: Yeah, you can't just throw this out. There *are* molecules with atom spin set.

I'd be OK with something that checks if there are any atoms with spin multiplicity set and performing your calculation if nothing is a radical.

ghutchis:
Handled by #1592 - thanks!

PR #1567, opened 2017-05-14
Fix regression + minor cppcheck report

serval2412:
regression from 624abd353f8ea2c889edb663adf84e50ee81e8d2 (2008-03-06)
- if (atoi(nbr->GetType()) == 77)
+ if (type == 77)
whereas:
+ int type = atoi(atom->GetType());

cppcheck:
[src/formats/mol2format.cpp:49]: (style) Statements following return, break, continue, goto or throw will never be executed

### DIFF
--- a/src/forcefields/forcefieldmmff94.cpp
+++ b/src/forcefields/forcefieldmmff94.cpp
@@ -3801,7 +3801,7 @@ namespace OpenBabel
             }
           }
 
-          if (type == 77)
+          if (atoi(nbr->GetType()) == 77)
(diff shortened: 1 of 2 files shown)


serval2412:
About forcefieldmmff94 part, I found this thanks to cppcheck report:
[src/forcefields/forcefieldmmff94.cpp:3804]: (style) Condition 'type==77' is always false
(Indeed at this line we're in a block "if type == 32")

ghutchis:
Thanks!

PR #1568, opened 2017-05-17
Convert tabs to spaces in testpdbformat.py

adamjstewart:
I'm trying to write a package for Open Babel for the Spack package manager (see https://github.com/LLNL/spack/pull/4256). The installation seems to work, but one of the unit tests crashes:
```
        Start 157: pytest_pdbformat
157/160 Test #157: pytest_pdbformat .................***Failed    0.04 sec
/blues/gpfs/home/software/spack-0.10.0/var/spack/stage/openbabel-2.4.1-g7kl4coxcxba2x2ugjpqbeipuembkpk7/openbabel-2.4.1/scripts/python:/scratch/ajstewart/spack-stage/spack-stage-z0O89Y/openbabel-2.4.1/spack-build/lib
  File "/blues/gpfs/home/software/spack-0.10.0/var/spack/stage/openbabel-2.4.1-g7kl4coxcxba2x2ugjpqbeipuembkpk7/openbabel-2.4.1/test/testpdbformat.py", line 32
    self.entryPDBwithInsertioncodes="""ATOM    406  N   VAL L  29      58.041  17.797  48.254  1.00  0.00           N  
                                                                                                                      ^
TabError: inconsistent use of tabs and spaces in indentation

CMake Error at pytest_pdbformat.cmake:20 (MESSAGE):
  1
```
The problem is that this file contains mixed tabs and spaces. This PR converts all tabs to spaces and allows the unit test to pass.

### DIFF
--- a/test/testpdbformat.py
+++ b/test/testpdbformat.py
@@ -24,12 +24,12 @@ class TestPDBFormat(BaseTest):
 
     def testInsertionCodes(self):
         """
-	Testing a PDB entry with insertion codes to distinguish residues
-	upon conversion to FASTA.
+        Testing a PDB entry with insertion codes to distinguish residues
+        upon conversion to FASTA.
         """
         self.canFindExecutable("babel")
 
-	self.entryPDBwithInsertioncodes="""ATOM    406  N   VAL L  29      58.041  17.797  48.254  1.00  0.00           N  
+        self.entryPDBwithInsertioncodes="""ATOM    406  N   VAL L  29      58.041  17.797  48.254  1.00  0.00           N  
 ATOM    407  CA  VAL L  29      57.124  18.088  47.170  1.00  0.00           C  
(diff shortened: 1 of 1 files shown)


baoilleach:
Thanks. Weird that it passes for me, but clearly wrong.

adamjstewart:
It could be a difference in the Python versions we are testing. Python 3 seems to be more strict about tabs vs. spaces than Python 2 was. According to http://stackoverflow.com/a/28326024/5828163:

> Python 3 disallows mixing the use of tabs and spaces for indentation.
>
> Python 2 code indented with a mixture of tabs and spaces should be converted to using spaces exclusively.
>
> When invoking the Python 2 command line interpreter with the -t option, it issues warnings about code that illegally mixes tabs and spaces. When using -tt these warnings become errors. These options are highly recommended!

Issue #1574, opened 2017-05-23
GUI does not build on Fedora 25


susilehtola:
Steps to reproduce:
1. Check out repository
2. mkdir objdir && cd objdir
3. cmake -DCMAKE_INSTALL_PREFIX=$(pwd)/../install ..
4. make -j9 install VERBOSE=1

Expected result:
openbabel builds and installs.

Actual result:
Build fails with
```
[ 86%] Linking CXX executable ../../bin/obgui
cd /home/work/openbabel/objdir/src/GUI && /usr/bin/cmake -E cmake_link_script CMakeFiles/obgui.dir/link.txt --verbose=1
/usr/lib64/ccache/c++    -pthread -O2 -DNDEBUG -O3 -DNDEBUG  -Wl,--enable-new-dtags -Wl,--fatal-warnings -Wl,--no-undefined -lc -Wl,--enable-new-dtags -Wl,--fatal-warnings -Wl,--no-undefined -lc  CMakeFiles/obgui.dir/OBGUI.o CMakeFiles/obgui.dir/stdwx.o CMakeFiles/obgui.dir/optswx.o  -o ../../bin/obgui -rdynamic ../../lib/libopenbabel.so.5.0.0 -pthread -Wl,-z,relro -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -lwx_baseu-2.8 -lwx_gtk2u_core-2.8 -lwx_gtk2u_adv-2.8 -lm -ldl -lz -Wl,-rpath,/home/work/openbabel/objdir/lib: 
/usr/bin/ld: CMakeFiles/obgui.dir/OBGUI.o: relocation R_X86_64_32 against `_ZN16SelFormatsDialog13sm_eventTableE' can not be used when making a shared object; recompile with -fPIC
CMakeFiles/obgui.dir/OBGUI.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
src/GUI/CMakeFiles/obgui.dir/build.make:149: recipe for target 'bin/obgui' failed
make[2]: *** [bin/obgui] Error 1
```

edit: it does work if -DBUILD_SHARED=OFF (in this case GUI is not built).

baoilleach:
Which OS?

susilehtola:
Fedora 25, and yes.

baoilleach:
Does it fail elsewhere if you turn off building the GUI?

susilehtola:
If the GUI is turned off, the build succeeds.

[...]

PR #1575, opened 2017-05-23
Compile position independent code for shared libraries.

susilehtola:
This PR adds the missing CMake rule about position independent object files, which are necessary to be able to compile shared libraries, see issue #1574 .


baoilleach:
I'd be okay with this as an option, but I'd like to get to the bottom of what's happening with #1574 first.

baoilleach:
Can you change this to a option for non-Windows builds and default it to true?

baoilleach:
On second thoughts, let's just merge it for now, and see whether it causes any problems. Any problems with this @ghutchis ?

ghutchis:
I'm fine with merging.

cowsandmilk:
The thing that confuses me is that [cmake sets this to be true](https://cmake.org/cmake/help/v3.0/prop_tgt/POSITION_INDEPENDENT_CODE.html) when build_type is SHARED , which it is here.

This is why [Fedora's build logs](https://kojipkgs.fedoraproject.org//packages/openbabel/2.4.1/5.fc26/data/logs/x86_64/build.log) already have -fPIC and work fine. And builds on my ubuntu system use -fPIC and work fine.

ghutchis:
Agreed 110% - it should be fine. But I don't see the harm in turning on -fPIC if there's an issue with CMake. I see this patch as a workaround, since there's clearly an issue.

baoilleach:
It's the static inclusion of external libraries into the gui executable that's the problem (I think?), not so much shared library build. The variable names in the CMake file are a bit confused.

Issue #249, opened 2016-01-18
ionic aluminum trichloride gets hydrides added


jeffjanes:
In both 2.3.2 and in git HEAD, the ionic form of aluminum trichloride gets mangled upon smiles round-tripping.  Hydrides are added.

/usr/bin/obabel -ismi <(echo '[Al+3].[Cl-].[Cl-].[Cl-]') -ocan
[AlH3+3].[Cl-].[Cl-].[Cl-]

ghutchis:
My initial thought was "oh, this is easy.. add in an ionic [Al+3] pattern to the implicit valences"

It's a bit harder than that.. In principal, you should check to make sure there's actually enough counter ions to balance..

I'd be OK with adding a small patch to the implicit valence rules, since I suspect the impact would be small.

jeffjanes:
Aren't the square brackets around an atom supposed to prevent any implicit valences from being assumed?  Is this a divergence between Daylight SMILES and OpenBabel SMILES?

baoilleach:
It's a bug, or rather it was never implemented. I have code for
implementing the smiles valence model lying around which would fix this, at
least when reading. Let me see if I can dig it up.

- Noel

baoilleach:
Just a note. It's a little weirder than I expected.

>obabel -:"[Al+3].C" -osmi
[AlH3+3].C

but
>obabel -:"[Al+3]" -osmi
[Al+3]

Haven't had a chance to track down the cause yet.

baoilleach:
The difference is due to perceived spin multiplicity on the Al:

Al impval 3 hyb 2 fcharge 3 spinmult 0
C impval 4 hyb 3 

Al impval 3 hyb 2 fcharge 3 spinmult 4

Next step is to track down why

baoilleach:
Ok - so I think it _is_ a problem with the SMILES implicit valence. It pops up in AssignSpinMultiplicity, where OBMols with a single atom shortcircuit the function, hence the differing behaviour in the two cases.

baoilleach:
Fixed by #1576.

PR #1576, opened 2017-05-23
Keep count of implicit hydrogens instead of inferring them

baoilleach:
This pull request makes two changes to the Open Babel API:
1. The count of implicit hydrogens on an atom is stored explicitly.
2. Kekulization has been rewritten as a perfect matching algorithm rather than dearomatization. Hat-tip to @johnmay for help with this.

Both of these changes address long-standing major deficiences of the toolkit.

Regarding (1), there have been many bug reports related to implicit hydrogens appearing/disappearing. Furthermore, it is really much simpler to do things this way, and the code will be faster as there will be no continual reperception of hydrogen counts. Previously there were several functions and flags to work-around the existing system such as ForceNoImplH; these all go away now.

Regarding (2), the previous approach was (IMO) solving the wrong problem. It is now much simpler and faster. Previously parts of the kekulization infrastructure were spread around several different files (some in SMILES format, some in OBMol, and some in kekulization.cpp) and had several associated trigger functions (e.g. IsKSingle).

So what's the downsides? Well, these are major changes and there are still some wrinkles:
1. The existing radical code was built as an adjustment on top of the old implicit valence code. There are no tests, and so some of this may not work now. I'll be meeting Chris Morley at an RSC meeting in June and will bend his ear.
2. Smiley Format doesn't read in SMILES correctly. I think this should be removed and the error messages and nice functions merged with the existing SMILES reader.
3. Since kekulization has been tightened up, we should implement the Daylight aromaticity model as a starting point for SMILES writing. Otherwise we will write out SMILES we cannot kekulize.
4. I implemented an implicit hydrogen adder based on typical valences (e.g. for PDB) with code from @dan2097. This is likely different from the old one (using IMPVAL values), and may need further tweaking.
5. I didn't know where to put global functions, e.g. OBKekulize and GetTypicalValences. Consider their current locations placeholders.
6. Now that I think of it, I scattered around a few TODOs. I should check those...

These are the things that I know about. Others may become apparent over time.


baoilleach:
The MSVC tests work fine for me, so I'm going to mute them on appveyor and then trigger a rebuild.

baoilleach:
Fixes #249, #1108, #1078, #1188, #1300, #628, #1179, #322, #1323, #1027, #1109, #298, #1136.

ghutchis:
Obviously, this is a great improvement overall and I'm willing to merge soon so we can find the corner cases, etc. (I also greatly appreciate that you've been cleaning out some of the old bugs.)

1. Radicals - I think for now, I'm willing to merge even if it breaks radicals, but I'd like at least some confirmation from @chrismorl that we can get radicals back before release.
2. I agree that we can merge Smiley with the existing reader. I believe @timvdm wanted to eventually replace the existing parser with Smiley, but I think a good first step is to integrate the useful bits.
3. Yes, we should implement the Daylight aromaticity model for SMILES.
4. Implicit hydrogen addition is *always* a pain, so the earlier we get this out, the fewer corner cases we'll have before release. (I fully expect after release, we'll get more complaints and bugs.)

In short, let me know if you're ready to merge and I'll be happy to finish the review.

baoilleach:
I agree with all of this, and I'd like to merge as soon as possible to winkle out any remaining issues asap and to draw a line under this part. I'll see @chrismorl on 22 June at a meeting, and I'll look into the issues surrounding radicals in the run up to that (I don't think it's really going to be a problem).

Appveyor is being annoying though. I'll have to skip another test - there's something not right about their system.

chrismorl:
I also don't think there is likely to be a problem with identifying 
radicals.  I'll try to look at the modified code before 22 June, but 
don't depend on this.

Chris

baoilleach:
Bump. Ok to merge?

ghutchis:
You should make an announcement on the mailing list.

ghutchis:
(I'm currently on a family holiday - back on Tuesday night or Wednesday morning.)

baoilleach:
For the record, also fixes #1343, #1320, #1207, #1166.

PR #1579, opened 2017-05-29
UFF: Fix conversion constant

aandi:
Probably a typo, the constant 2G=664.12 (not 644...) converts units to kcal/mol, see also equation (13)f from the original paper by Rappé et al. or e.g. the RDKit (AngleBend.cpp, look for beta = ...) source. I also verified the outcome of the corrected version, the code now produces a stiffness of 105.5 kcal/(mol rad^2) (or rather 210.976371589 * 0.5 = 105.488185794) for a C_3-N_R-C_R amide angle, which corresponds to the example given in Rappé et al.


ghutchis:
Thanks!

Issue #1584, opened 2017-06-02
Depiction of multifragment mol-file causes overlaps


fredrikw:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: Trunk (with addition of PR #1582)
Operating system and version: OSX 12.5

## Expected Behavior
Depiction of a molfile containing ZnCl2 ([Zn+2].[Cl-].[Cl-]) (content of molfile:

  ChemDraw06021710362D

  3  0  0  0  0  0  0  0  0  0999 V2000
   -0.2736    0.3162    0.0000 Cl  0  5  0  0  0  0  0  0  0  0  0  0
    0.2736    0.3162    0.0000 Zn  0  2  0  0  0  0  0  0  0  0  0  0
    0.1911   -0.3162    0.0000 Cl  0  5  0  0  0  0  0  0  0  0  0  0
M  CHG  3   1  -1   2   2   3  -1
M  END
) should keep the coordinates from the file in the resulting image.
![zncl2_chemdraw](https://cloud.githubusercontent.com/assets/86855/26718254/ab9aa108-4780-11e7-88ea-36e0949cd7df.png)

## Actual Behavior
The resulting image gets all atoms overlayed.
![zncl2_openbabel](https://cloud.githubusercontent.com/assets/86855/26718289/ce5157aa-4780-11e7-91f3-05b1aa8180e7.png)

## Steps to Reproduce
1. obabel -imol -O ~/Desktop/ZnCl2_openbabel.png
2. paste the above file content
3. look at resulting image

fredrikw:
I found the reason and the problem only occured when there were no bonds in the molecule. The current scaling was based on average bond length and set to 1 if no bonds were present. That meant that depending on the coordinates used in the original file, the fragments could be almost on top of each other (or separated by a huge amount). I created a PR (#1585) to use the minimum atom distance instead of average bond length. There might be a better option, but this seems to work.

baoilleach:
Fixed by #1585.

PR #1585, opened 2017-06-02
Fix multifragment depiction

fredrikw:
A fix for #1584. Molecules are normally scaled by the average bond length, but for molecules without bonds (e.g. inorganic salts) the scale was always set to 1. This PR calculates the minimal distance between two atoms and uses that instead of average bond length. There might be better ways, but this fixed my issue.

### DIFF
--- a/src/depict/depict.cpp
+++ b/src/depict/depict.cpp
@@ -388,23 +388,27 @@ namespace OpenBabel
       double f;
       if(mol->NumBonds()>0) {
         f = d->bondLength / averageBondLength;
-      } else {
-        f = d->bondLength;
+      } else if(mol->NumAtoms()>1){
+        /* In molecules without bonds but more than one atom,
+           use the minimum atom distance as a substitute for average bond length. */
+        f = DBL_MAX;
         OBAtomIterator i2;
         OBAtom *atom2;
+        vector3 a1pos;
+        double currdist;
         for (atom = d->mol->BeginAtom(i); atom; atom = d->mol->NextAtom(i)) {
+          a1pos = atom->GetVector();
           for (atom2 = d->mol->BeginAtom(i2); atom2; atom2 = d->mol->NextAtom(i2)) {
             if(atom != atom2) {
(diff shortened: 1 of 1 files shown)


baoilleach:
Can't you only do this when no bonds are present? Otherwise, you slow down the general case.

baoilleach:
To speed things up, don't calculate the square root until after the two loops. That is, use the square of the distance. It's an old trick but a good one. Also can you include in the code a comment as to what/why it's doing?

baoilleach:
To speed things up, don't calculate the square root until after the two loops. That is, use the square of the distance. It's an old trick but a good one. Also can you include in the code a comment as to what/why it's doing?

...oh, and welcome to the team. :-) Great to see pull requests from you.

baoilleach:
Is this also a fix for #1317?

fredrikw:
I will move the square root tomorrow. I considered leaving it out from the beginning, but I guessed that there are seldom molecules with no bonds and more than a few atoms in them so that the loop won't be called that much... Do you think that the distance calculations should be changed and cached as well? (Instead of calling `atom->GetX() - atom2->GetX()` twice) I'm not enough of a C++-programmer to tell what level of optimization is best.

 Regarding #1317, I think that the actual error mentioned there (
```
*** Open Babel Error  in SVGFormat
  Molecule  needs 2D coordinates to display in SVGformat
```
) has been fixed for a while (at least I don't see it when I run the example code on a version from before my changes) but with this change you will also get a good looking image instead of 
test_1317
 :-)

baoilleach:
Better still would be to store the 'vector3' of coordinates from the first atom in a variable, and then call its distSq method to get the squared distance to the vector3 of the second atom.

Also, while we are micro-optimising :-), you could replace your std::min with a basic "if (newdist < mindist_so_far) mindist_so_dar = newdist;". Using std::min does a comparison and then an assignment. The simpler "if statement" only does the assignment sometimes.

fredrikw:
I incorporated the suggested changes and it still seems to work :-)

PR #1586, opened 2017-06-09
Introduce std:isnan for older versions of MSVC

mwojcikowski:
Define isnan for older versions of MSVC. Kudos to Christoph Lassner for his blog post: https://www.christophlassner.de/collection-of-msvc-gcc-compatibility-tricks.html

### DIFF
--- a/src/conformersearch.cpp
+++ b/src/conformersearch.cpp
@@ -21,6 +21,15 @@ GNU General Public License for more details.
 #include <openbabel/math/align.h>
 #include <openbabel/forcefield.h>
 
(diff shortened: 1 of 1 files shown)


baoilleach:
Our usual approach is not to add things to the standard namespace but to have a level of indirection.

That is, have a variable ISNAN which we use in the code, and assign a value to it at the top, either std::isnan or your templated definition.

baoilleach:
I've sent you a pull request for review...

mwojcikowski:
LGMT, sorry for the delay.

PR #1593, opened 2017-06-27
Change silicon to correct MM3 atom type

keipertk:
Silicon is incorrectly assigned as Magnesium. 

It should be replaced with the only silicon type in the mm3 parameter list - #19 silane
https://dasher.wustl.edu/tinker/distribution/params/mm3.prm


ghutchis:
Great - looks like a copy/paste error. Thanks!

Issue #1591, opened 2017-06-21
Unsuccessful compilation of the library: problem with dlhandler_unix.cpp


axvutt:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 2-4-1
Operating system and version: Window 10, 64 bit

## Expected Behavior
Successful build of the Open Babel library.

## Actual Behavior
* cmake:
C:\OpenBabel\build>cmake ..\openbabel-2-4-1 -G"MinGW Makefiles" -DEIGEN2_INCLUDE_DIR=C:\Eigen-3.2.9 -DBUILD_GUI=OFF
--> Operation succeeded (did not find Eigen though...)

* mingw32-make:
C:\OpenBabel\build>mingw32-make UNICODE=0 BUILD=release SHARED=1 MONOLITHIC=1
--> Compilation error:

> C:\OpenBabel\openbabel-2-4-1\src\dlhandler_unix.cpp: In static member function 'static bool DLHandler::openLib(const string&)':
> C:\OpenBabel\openbabel-2-4-1\src\dlhandler_unix.cpp:156:36: error: cannot convert 'const char*' to 'LPCWSTR {aka const wchar_t*}' for argument '1' to 'HINSTANCE__* LoadLibraryW(LPCWSTR)'
>      if(LoadLibrary(lib_name.c_str()))
>                                     ^
> src\CMakeFiles\openbabel.dir\build.make:1387: recipe for target 'src/CMakeFiles/openbabel.dir/dlhandler_unix.obj' failed
> mingw32-make[2]: *** [src/CMakeFiles/openbabel.dir/dlhandler_unix.obj] Error 1
> CMakeFiles\Makefile2:1078: recipe for target 'src/CMakeFiles/openbabel.dir/all' failed
> mingw32-make[1]: *** [src/CMakeFiles/openbabel.dir/all] Error 2
> Makefile:139: recipe for target 'all' failed
> mingw32-make: *** [all] Error 2

IMHO, the error is caused by LoadLibrary() which is replaced by LoadLibraryW() when UNICODE is defined, whatever its value (see 'winbase.h', lines 2232, 2375 and 2582). Its argument, in 'dlhandler_unix.cpp' is not implicitely converted, hence the error.
Suggested fix: replace LoadLibrary() by LoadLibraryA().

baoilleach:
Does it work if mingw32-make is used without any additional options?

axvutt:
[...]

PR #1594, opened 2017-06-28
Fix 'dlhandler_unix.cpp' UNICODE bug

axvutt:
Fix suggested in issue #1591.

lib_name.c_str() alwarys returns a const char*. (C string)
The definition of LoadLibrary() changes depending on whether the macro UNICODE is defined or not (see 'winbase.h'). In the former case, the LoadLibrary() is actually replaced by LoadLibraryW() which accepts LPCWSTR {aka const wchar_t*} type, and the compilation fails due to mismatching data type and non-implicit conversion. In the latter, LoadLibraryA() is called. That version works with char*.

The best solution seems to be the one detailed in the following article:
http://www.cplusplus.com/forum/articles/16820/
the WinAPI provides the macros TEXT(), _T() and T() which do the same thing: convert char* to LPCWSTR and vice versa depending on the definition (or non-definition) of the UNICODE macro.

### DIFF
--- a/src/dlhandler_unix.cpp
+++ b/src/dlhandler_unix.cpp
@@ -153,7 +153,7 @@ int DLHandler::findFiles (std::vector<std::string>& file_list,
 bool DLHandler :: openLib(const string& lib_name)
 {
 
-    if(LoadLibrary(TEXT(lib_name.c_str())))
+    if(LoadLibrary( (PTCHAR) lib_name.c_str()))
         return true;
 
     unsigned long err = GetLastError();
--- a/src/dlhandler_unix.cpp
+++ b/src/dlhandler_unix.cpp
@@ -153,7 +153,7 @@ int DLHandler::findFiles (std::vector<std::string>& file_list,
 bool DLHandler :: openLib(const string& lib_name)
 {
 
-    if(LoadLibrary(lib_name.c_str()))
+    if(LoadLibrary(TEXT(lib_name.c_str())))
         return true;
 
     unsigned long err = GetLastError();


baoilleach:
I'm afraid that this fix doesn't work on Windows 7. If you look in winnt.h what is __TEXT defined as?

I have:
#define __TEXT(quote) L##quote      // r_winnt
#define TEXT(quote) __TEXT(quote)   // r_winnt

...so all TEXT does is stick the letter L in front of the string, which doesn't do anything sensible if it's a variable name.

baoilleach:
PTCHAR works for me though. Could you try that?

axvutt:
Aaaaah my bad! It's indeed a macro, not a function...

>  Put all string **literals** in _T() macros.

so not variables ^^'. The definition I have is similar to yours:
```
#define __TEXT(q) L##q
#else
#define __TEXT(q) q
#endif
\\ comment
#define TEXT(q) __TEXT(q)
```

Indeed, in that same article,they say:

> `TCHAR` is #defined as either a `char` or a `wchar_t` depending on whether or not the `UNICODE` macro was defined.

And, as you suggest, writing
```
if(LoadLibrary( (PTCHAR) lib_name.c_str() ))
        return true;
```
worked fine for me. However, now I have a doubt when they say:

> Never ever ever C style cast C strings from one type to another. C style casts mask very important compiler errors. Avoid C++ style casts also. Basically if you're getting type errors with your strings -- it's because you're doing something wrong. Don't try to cast around the problem.

So this might not be exactly what we want...

On the other hand, there is something I do not get... This `openLib()` method is implemented in the same way in both `dlhandler_win32.cpp` and `dlhandler_unix.cpp` when compiled with MinGW (in the latter, it is surrounded by `#ifdef __MINGW32__` ... `#else`). The compile error came only from `dlhandler_unix.cpp`. Is it because of some different `#include`s?

baoilleach:
Given that the MinGW build still fails further along, this cannot be tested to check whether it actually works. So, let's put this on-hold until all of the MinGW problems are fixed. I'll try and find time to figure it out...

baoilleach:
Can you try the following change and see whether the build completes for you too?

--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
     CACHE STRING "Binary build output directory")
 endif()
 if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
-  if(UNIX)
+  if(UNIX OR MINGW)
     set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}"
       CACHE STRING "Library build output directory")
   else() # This is preferable on WIndows for example - bin = lib directory.

axvutt:
> Given that the MinGW build still fails further along, this cannot be tested to check whether it actually works.

Sorry, I didn't understand this. The change
```
- if(LoadLibrary( lib_name.c_str() ))
+ if(LoadLibrary( (PTCHAR) lib_name.c_str() ))
```
did work on my machine.

> Can you try the following change and see whether the build completes for you too?
> 
> ```
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -88,7 +88,7 @@ if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
>      CACHE STRING "Binary build output directory")
>  endif()
>  if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
> -  if(UNIX)
> +  if(UNIX OR MINGW)
>      set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}"
>        CACHE STRING "Library build output directory")
>    else() # This is preferable on WIndows for example - bin = lib directory.
> ```


The change from `if(UNIX)` to `if(UNIX OR MINGW)`, keeping the original `dlhandler_unix.cpp`, does not work in my case. It still tries to compile `dlhandler_unix.cpp` without success, due to the same error.

baoilleach:
Are you saying that you can build the entirety of Open Babel, and for example, run the test (via "make tests") after the LoadLibrary change? I also had to do the second step.

axvutt:
I think I have not been specific enough, sorry for that.
No, not entirely indeed. In my last attemps I was building only `openbabel` with
```
mingw32-make openbabel
```
because, I thought, the problem comes from there. The latter command successfully (or, should I say more correctly "without errors") builds what it has to build with the `(PTCHAR)` change .

Actually, I am not using any of the optional libraries: Eigen 3.0 (even though I have it), libxml2, zlib, wxWidgets 2.8 (even though I have it, but I don't want the GUI). In fact, I just want to use the C++ library API for my in-house developments...

Now, I tried the following (still with the (PTCHAR) change):
```
C:\OpenBabel\build2>cmake ..\openbabel-2-4-1 -G"MinGW Makefiles" -DBUILD_GUI=OFF
```
the displayed information is enclosed in [cmake.txt](https://github.com/openbabel/openbabel/files/1116542/cmake.txt).
```
C:\OpenBabel\build2>mingw32-make -j4
```
this builds all it has to build without giving errors.
```
C:\OpenBabel\build2>mingw32-make test
```
with only 34% tests passed, 95 tests failed out of 144. The details are in the enclosed [tests.txt](https://github.com/openbabel/openbabel/files/1116543/tests.txt). (I had to shut test 124 down manually because it was stuck there for very long, see elapsed time).

I have the feeling I am doing it totally wrong... :(

PR #1596, opened 2017-07-05
Fix pubchem JSON handling of enum types as ints

mcs07:
At some point in the past, it seems that the PubChem JSON format returned by the PubChem REST API changed to return enum type names as integers instead of the string representation. So elements are atomic numbers instead of symbols, bond orders are integers instead of "single", "double", etc.

[Discussion on openbabel-devel mailing list](https://sourceforge.net/p/openbabel/mailman/message/35918486/)

### DIFF
--- a/src/formats/json/pubchemjsonformat.cpp
+++ b/src/formats/json/pubchemjsonformat.cpp
@@ -166,15 +166,14 @@ class PubChemJSONFormat : public OBMoleculeFormat
     Json::Value elements = molRoot["atoms"]["element"];
     pmol->ReserveAtoms(eAids.size());
     for(Json::ArrayIndex i = 0; i < eAids.size(); i++) {
-      if (eAids[i].isInt() && elements[i].isString()) {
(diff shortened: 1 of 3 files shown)


ghutchis:
While it's clearly not a widely used format, I'm hesitant to merge a change that will now fail on old text-based responses. 

Shouldn't you check if the response is non-numeric and handle with the existing code and then handle with the new code if it's a number? That seems the best solution, IMHO.

mcs07:
Yeah, makes sense, just a little more work!

I'll wait until Noel's new element handling code is merged first, otherwise there will be a clash.

baoilleach:
....and it would be great to have a test case (e.g. in testbindings.py) for
each of the formats supported.

Actually, I was going to suggest merging yours first. But either way :-)

baoilleach:
Bump...(element handling merged)

mcs07:
I think this is now ready for review

PR #1599, opened 2017-07-10
Correct MM3 carboxyl oxygen atom type definition

keipertk:
MM3 has an atom type definition specifically for carboxyl oxygens. The present code defines carboxyl oxygens as carboxylate ions (which remains the best choice when using MM2).


ghutchis:
I think my main concern with this change is that you don't address a case for 47, e.g. C(=O)O-

I agree it's probably a good change for many cases, though.

PR #1608, opened 2017-08-01
 Fix some small misspellings in the csharp bindings

cmanion:
the functions for OBTorsionIter and OBRingIter are misspelled in the swig interface file for the csharp bindings

### DIFF
--- a/scripts/openbabel-csharp.i
+++ b/scripts/openbabel-csharp.i
@@ -132,7 +132,13 @@
 		get{return new OBVector3(0,0,0);}
 	}
 %}
-  
+%typemap(cscode) OpenBabel::matrix3x3
+%{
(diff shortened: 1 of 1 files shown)


baoilleach:
Hi @cmanion. Thanks for these fixes, additions. It's a good idea to put commits addressing separate issues in different pull requests / branches so that they can be easily reviewed separately.  For more info, see https://guides.github.com/introduction/flow/

baoilleach:
The C++ library already has matrix multiplication. You need to provide some explanation or justification for your pull request.

cmanion:
So I am new to github and did not mean to commit the matrix multiplication code. However, currently matrix multiplication operators are not wrapped with the C# bindings. So C# code was added in the swig bindings as was done with the vector math operators, which weren't wrapped either.

PR #1614, opened 2017-08-15
Simplify user installation by creating build directory upfront

zwanzigpl:
Build dir contains empty .keep file just to avoid creating it by user. Any other build files will be ignored


ghutchis:
(shrug) - I don't see how this significantly improves things. Most people install from a binary package from a package manager or download the Windows binary.

PR #1601, opened 2017-07-13
Improve performance of element handling

baoilleach:
As described over at https://github.com/openbabel/enhancement-proposals/pull/4.

Briefly, the old element handling code had some performance problems. I have written a faster replacement. It has the following additional benefits:
1. No need for a global etab or isotab and associated initialization.
2. The elements are now available as consts (e.g. OBElements::Carbon) to replace functions such as IsCarbon().
3. No need for text/header files in data dir.

Timing comparison: get the symbol for elements 1 to 100 and convert back to atomic number (repeat x10000). 0.82s old vs 0.02s new. A more realistic conversion is to just do carbon: 0.14s old vs 0.01s new.

```
void test_OBElements()
{
  clock_t start = clock();
  unsigned int tot = 0;
  for (unsigned int N = 0; N < 10000; ++N) {
    for (unsigned int i = 1; i < 100; ++i) {
      const char *elem = OBElements::GetSymbol(6);
      unsigned int atomic_num = OBElements::GetAtomicNum(elem);
      tot += atomic_num;
    }
  }
  clock_t stop = clock();
  double sum = (double)(stop - start) / CLOCKS_PER_SEC;
  printf("%.2f s\n", sum);
  
  OBElementTable etab;
  start = clock();
  tot = 0;
  for (unsigned int N = 0; N < 10000; ++N) {
    for (unsigned int i = 1; i < 100; ++i) {
      const char *elem = etab.GetSymbol(6);
      unsigned int atomic_num = etab.GetAtomicNum(elem);
      tot += atomic_num;
    }
  }
  stop = clock();
  sum = (double)(stop - start) / CLOCKS_PER_SEC;
  printf("%.2f s\n", sum);
}
```

PR #1617, opened 2017-08-21
Fix install regression with element data

bbucior:
Fixes a minor installation regression from PR #1601.  `make install` crashed with the following error:

```
CMake Error at data/cmake_install.cmake:39 (file):
  file INSTALL cannot find
  "/cygdrive/c/Users/Benjamin/Git/Contrib/openbabel/data/element.txt".
Call Stack (most recent call first):
  cmake_install.cmake:68 (include)
```

Removing the deleted files from the cmake target allowed installation to proceed.

I haven't used Travis much yet, but would a `make install` test in CI be useful?

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -5,7 +5,6 @@ set(to_install
     atomtyp.txt
     babel_povray3.inc
     bondtyp.txt
-    element.txt
     eem.txt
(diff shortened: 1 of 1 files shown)


ghutchis:
Yes, clearly a "make install" target for Travis would be useful - it would have caught this one.

PR #1622, opened 2017-08-28
Added some missing formats to the static build

psavery:
formats.cmake is used for static builds, but it does
not look entirely up-to-date with the CMakeLists.txt in the
same directory. This PR adds several formats that are present
in CMakeLists.txt but missing in formats.cmake.


ghutchis:
Actually, would you mind taking a few minutes and making it consistent - making *everything* go through formats.cmake?

I think the maintenance problem is that people need to add formats in both places.

psavery:
Okay, I tried out something new.

I noticed that the top part of CMakeLists.txt and formats.cmake are almost identical. To avoid having people re-write the formats that they add into both (and consequently avoid the problem of people forgetting to do so), I copied over the top portion of CMakeLists.txt to formats.cmake, and now static and dynamic builds both include formats.cmake.

There were different instructions for shared vs static builds for only a few formats. For Inchi, I just wrapped the part that was in the shared build with a "if(BUILD_SHARED)" and the static part in the "elseif(WITH_STATIC_INCHI)" section. For LIBXML2, they were identical, so I just made sure that it runs when the proper conditions are set. For GenBankFormat, that one was only built statically (this is because it uses a function from the fasta format, and it can't build dynamically because it doesn't link to it). So I just wrapped it in "if(NOT BUILD_SHARED)".

What do you think?

I tested this for both dynamic and shared builds, and I was able to convert formats just fine with the resulting obabel executables (I tried cml, inchi, and xyz for both).

psavery:
If you'd rather me do it the other way, just let me know and I'll go back keep the formats.cmake as static-only.

ghutchis:
I think the GenBank issue is separate - but I haven't heard many complaints about it not being available.

psavery:
Thanks Geoff! 

Yeah, if GenBank becomes a problem in the future, I think we can do one of these:

a) Copy and paste the fasta function it references into GenBankFormat itself (the function is not a member function of the Fasta format)
b) Link GenBankFormat with the Fasta .so on dynamic builds
c) Put the fasta function it references somewhere generic so both GenBankFormat and Fasta can reference it.

PR #1623, opened 2017-08-28
In SiestaFormat, print warnings to cerr

psavery:


ghutchis:
I'm fine with this, but I'd love to see this wrapped to be consistent with other formats.

obErrorLog.ThrowError(__FUNCTION__, errorMsg.str() , obWarning);

PR #1626, opened 2017-08-30
Handle undervalent atoms and radicals in Mol files and Smiles

baoilleach:
This is a follow-on from the work on implicit hydrogens. In the end no changes were made to the data structures used to store information, but the handling of the existing file formats has been tweaked.

Storing undervalent atoms in MOL files now uses the valence field, rather than M RAD. M RAD is only used if a spin has been specified.

Now reading undervalent atoms from SMILES does not set the spin (it did previously). For depiction, changes were needed to ensure that radical dots were depicted in the same situations as before (by testing for either spin or undervalence).

### DIFF
--- a/src/depict/depict.cpp
+++ b/src/depict/depict.cpp
@@ -28,6 +28,7 @@ GNU General Public License for more details.
 #include <openbabel/stereo/stereo.h>
 #include <openbabel/stereo/cistrans.h>
 #include <openbabel/obiter.h>
+#include <openbabel/obfunctions.h>
 
 #include <cmath>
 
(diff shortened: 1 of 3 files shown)


baoilleach:
I missed a couple of things in my notes above...

Since we needed this functionality anyway for undervalent atoms, I also added a MOL-file write option to force setting the valence field on every atom. If this option is used, then any reader (that honors it) will not apply an implicit valence model but just read off the number of hydrogens directly from the valence field. Potentially useful if you do not trust the reader to add the correct number of hydrogens.

Although I discussed this in the original proposal, I didn't add an 'op' to set the spin state of undervalent atoms. It could be done of course, but the syntax was sufficiently involved (how to specify which atoms?, what default atoms?) that I'd rather wait until someone actually has a use case.

ghutchis:
The code looks fine. I think it's worth testing against ChemDraw and ChemDoodle import/export of Molfiles, but that can come later.

Issue #1624, opened 2017-08-29
math parsing issue during compile


Justinzobel:
## Environment Information
Open Babel version: 2.4.0/1
Operating system and version: Linux (Solus 3)

## Expected Behavior
Compile

## Issue
cmake '-DCMAKE_CFLAGS=-mtune=generic -march=x86-64 -g2 -O2 -pipe -fPIC -Wformat -Wformat-security -fno-omit-frame-pointer -fexceptions -D_FORTIFY_SOURCE=2 -fstack-protector --param ssp-buffer-size=32 -fasynchronous-unwind-tables -ftree-vectorize -feliminate-unused-debug-types -Wall -Wno-error -Wp,-D_REENTRANT' '-DCMAKE_CXX_FLAGS=-mtune=generic -march=x86-64 -g2 -O2 -pipe -fPIC -fno-omit-frame-pointer -fexceptions -D_FORTIFY_SOURCE=2 -fstack-protector --param ssp-buffer-size=32 -fasynchronous-unwind-tables -ftree-vectorize -feliminate-unused-debug-types -Wall -Wno-error -Wp,-D_REENTRANT' '-DCMAKE_LD_FLAGS=-Wl,--copy-dt-needed-entries -Wl,-O1 -Wl,-z,relro -Wl,-z,now -Wl,-z,max-page-size=0x1000' -DCMAKE_LIB_SUFFIX=64 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr .
-- The C compiler identification is GNU 6.4.0
-- The CXX compiler identification is GNU 6.4.0
-- Check for working C compiler: /usr/lib64/ccache/bin/x86_64-solus-linux-gcc
-- Check for working C compiler: /usr/lib64/ccache/bin/x86_64-solus-linux-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib64/ccache/bin/x86_64-solus-linux-g++
-- Check for working CXX compiler: /usr/lib64/ccache/bin/x86_64-solus-linux-g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Using included inchi library.
[...]

PR #1630, opened 2017-09-04
Compiler version parsing and comparison from CMake 2.8

cowsandmilk:
Fix for #1624 by using CMake's version parsing. Also move to using
CMake's internal version comparison operators.


baoilleach:
Looks good.

PR #1638, opened 2017-09-13
Implement the Daylight aromaticity model as described by John Mayfield

baoilleach:
John has described the atom contributions in the Daylight aromaticity plus some extensions (e.g. [S+]-[O-] treated the same as [S]=[O].) https://figshare.com/articles/Daylight_Aromatic_Atoms/3370945

I have implemented these in the code. This replaces the existing code which aromatised things at least some of which shouldn't have been, at least in the context of SMILES strings. For example, rings containing radicals. This is partly because the existing code couldn't test for exact valence, as the hydrogens were floating and so a radical nitrogen looked similar to a nitrogen with a hydrogen.

We may support multiple aromaticity models in future, but at least now we have a firm foundation in a documented 'spec'.

I have a benchmark set of potentially aromatic rings over at https://github.com/nextmovesoftware/smilesreading and I have been comparing the before vs after picture. As far as I can tell, everything is an improvement.


ghutchis:
You'll need to regenerate the aromaticity test to pass Travis.

More importantly, there was some discussion about making sure we had "flavors" of aromaticity models - that is, there would be an OBDaylightAromaticityTyper and an OBXYZ AromaticityTyper, etc.

Does that seem like a reasonable goal? I'm fine if we say this is step one in that direction and other aromaticity models will be implemented as needed in the future.

baoilleach:
Ah yes - of course there are issues with the tests - that's what I get when I do a late-night pull request.

Regarding the flavors, for sure. It's just a different set of switch statements, or same switch statements with different electron contributions. What the API should be needs a small bit of thought, for example what's expected if you apply a different model and then write out a SMILES? Does the library need to remember which model has been applied? Could I write up a proposal separate to this pull request?

ghutchis:
Yes, I think flavors of aromaticity should be a separate enhancement discussion to get the API right.

I'll take a look at the pull request this weekend. Thanks.

baoilleach:
Actually, there may be another P atom type still to do... Maybe hold off for now.

baoilleach:
Sorry about that - I think it's ready now.

marcabrus:
Since I have a similar request I'll write it here. I'm experiencing some trouble when porting SMILES from openbabel standardization to other software, since SMILES notation could be not recognized due to a different aromaticity model implemented. 
Is it possible to add more than one model as option for smiles output?
For instance in chemaxon there are three different aromaticity models and sometimes you need to use the basic model (that doesn't take in account taoutomers or mesomers or consider any external double bond influencing aromaticity) in order to make the SMILES compatible to other software. 
Of course this influences also descriptors like number of aromatic bonds and is it impossible in some cases to have reproducibility.

baoilleach:
Hi there, I am currently studying the reasons behind problems that toolkits have reading SMILES from other toolkits, and just gave a talk on this topic at the ACS: https://www.slideshare.net/NextMoveSoftware/a-de-facto-standard-or-a-freeforall

In short, the problem is not the aromaticity model but rather bugs. If you can provide a specific example, I can look into it. Also I recommend you use the development version of Open Babel, where multiple issues have been addressed.

That said, I do intend to add support for multiple aromaticity models...but it won't solve the problem you are describing.

marcabrus:
I can address as example la teobromine:
image
Updated CDK, RDkit, or Openbabel works on a Daylight-like model so it finds quindi con 10 aromatic bonds e 2 aromatic rings while basic model like implemented in prediction software Vega-cdk1.4.9 (https://www.vegahub.eu/) - , in padel-cdk 1.4.15 (http://www.yapcwsoft.com/dd/padeldescriptor/)- and like basic option in marvinsketch) reads only 5 aromatic bonds. Of course this is related to the lack of update of several software, but sometimes you need to use also old software for your work.

baoilleach:
@johnmay (the CDK developer) tells me that Padel is using an old version of the CDK that doesn't read these strings correctly. Given that I don't know what subset of aromaticity Padel or Vega are able to read, it would be hard to have an aromaticity model that would allow them to read it. But why use aromatic SMILES at all? - just use Kekule SMILES - they should be able to read that and then apply their own aromaticity model.

PR #1652, opened 2017-10-04
Correct handling of stereo when writing InChIs

baoilleach:
It turns out that the part of the InChI writer that handles the stereo assumes that AtomIds are simply AtomIdxs-1, but this is not true in general, and can lead to an incorrect stereo layer. The fix is to properly convert from AtomIds to AtomIdxs.

Although this only affects 11 molecules in ChEMBL when converting straight from SMI to InChI (and so wouldn't have been noticed before), it was already causing my second RInChI test case to fail (a test case taken from the RInChI docs). My guess is that it must be copying the molecule when generating the shared pointer, thus rearranging either the Ids or Idxs.

To be honest, I'm not 100% convinced that we need AtomIds - I must talk to Tim about this sometime.

### DIFF
--- a/src/formats/inchiformat.cpp
+++ b/src/formats/inchiformat.cpp
@@ -253,6 +253,12 @@ int InChIFormat::SkipObjects(int n, OBConversion* pConv)
   return ifs.good() ? 1 : -1;
 }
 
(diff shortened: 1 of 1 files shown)


ghutchis:
I think Tim just really wanted to move away from "index from 1" - we don't need them.

PR #1655, opened 2017-10-05
Allow RSMI format to read partial reactions

baoilleach:
It is perfectly valid in RSMI to be missing a reactant or product (or indeed both). I've added support for these so that the following are now read without any trouble:
```
>>N
O>>
>>
```
These are also written out exactly as above - no changes were made to make this happen. Test cases included.


ghutchis:
Can you check the conflict?

baoilleach:
My on-line merge attempt failed. I'll sort this out on my PC asap.

baoilleach:
Closing in favor of #1660

PR #1656, opened 2017-10-05
Add support for agents to RXN file format

baoilleach:
BIOVIA have started supporting agents in RXN file format. This may be due to the fact that ChemAxon (and NextMove) have been doing it for a number of years.

I've added read and write support for agents. On write, there are a number of options so that if you need to deal with an old version of BIOVIA you can ignore the agents or treat them some other way.

The following should now work:
```
obabel -irsmi -:"C>N>O" -orxn
```
We should generate 2D coordinates for --gen2D (but we don't). This will come automatically if/when we move to an overlay for reactions, so I'm not going to worry about supporting it now.

### DIFF
--- a/src/formats/rxnformat.cpp
+++ b/src/formats/rxnformat.cpp
@@ -39,9 +39,17 @@ public:
   {
       return
         "MDL RXN format\n"
-        "The MDL reaction format is used to store information on chemical reactions.\n"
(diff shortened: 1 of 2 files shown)


ghutchis:
This will need to be re-opened. Evidently merging the various agent patches in different orders led to build failures.

I reverted this pull.

PR #1663, opened 2017-10-10
Fix reading atom symbols from XSF file

sencer:
This is from the [official definition of XSF format](http://www.xcrysden.org/doc/XSF.html#__toc__3):

> An entry for an atom looks like:
>  AtNum   X Y Z
> where AtNum stands for atomic number (or symbol), while X Y Z are
> Cartesian coordinates in ANGSTROMS units. Here is one example: # Please
> enter the commit message for your changes. Lines starting

### DIFF
--- a/src/formats/xsfformat.cpp
+++ b/src/formats/xsfformat.cpp
@@ -81,6 +81,7 @@ namespace OpenBabel
     vector<string> vs;
     vector<vector3> atomPositions;
     bool createdAtoms = false;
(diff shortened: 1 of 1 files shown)


baoilleach:
Thanks. Can you put that information either in a comment in the code, or add a test case (or both!).

sencer:
Hi, I just added a comment line. No tests for now, unfortunately.

PR #1665, opened 2017-10-12
Correct multiple swig version location for python binding compilation

jjgomera:
I use debian stretch and for get a python binding i've had problem to install from pip, installation always search for swig program and in my system the installed version is the swig3.0, so i configure the setup to search between both version to choose always any installed version.

### DIFF
--- a/scripts/python/setup.py
+++ b/scripts/python/setup.py
@@ -1,10 +1,11 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 import os
 import subprocess
 import sys
 from distutils.command.build import build
 from distutils.command.sdist import sdist
 from distutils.errors import DistutilsExecError
+from distutils.spawn import find_executable
 from distutils.version import StrictVersion
 from setuptools.command.build_ext import build_ext
 from setuptools.command.install import install
@@ -13,7 +14,7 @@ from setuptools import setup, Extension
 
 __author__ = 'Noel O\'Boyle'
 __email__ = 'openbabel-discuss@lists.sourceforge.net'
-__version__ = '2.4.0'
+__version__ = '2.4.1'
 __license__ = 'GPL'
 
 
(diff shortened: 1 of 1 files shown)


scripts/python/setup.py
```diff
@@ -1,10 +1,11 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
ghutchis: I don't want to make that change. If someone still has Python 2, or a Python binary "python" (e.g. my Python 3.6 binary) it will fail.

scripts/python/setup.py
```diff
@@ -52,15 +53,15 @@ def pkgconfig(package, option):
...
+        version = pkgconfig('openbabel', '--modversion')
         if not StrictVersion(version) >= StrictVersion('2.3.0'):
             print('Warning: Open Babel 2.3.0 or later is required. Your version (%s) may not be compatible.' % version)
-        include_dirs = pkgconfig('openbabel-2.0', '--variable=pkgincludedir')
+        include_dirs = pkgconfig('openbabel', '--variable=pkgincludedir')
         library_dirs = pkgconfig('openbabel-2.0', '--variable=libdir')
         print('Open Babel location automatically determined by pkg-config:')
     except PkgConfigError as e:
         print('Warning: %s.\nGuessing Open Babel location:' % e)
-        include_dirs = '/usr/local/include/openbabel-2.0'
```
ghutchis: This is preferred.

scripts/python/setup.py
```diff
@@ -52,15 +53,15 @@ def pkgconfig(package, option):
 def locate_ob():
     """Try use pkgconfig to locate Open Babel, otherwise guess default location."""
     try:
-        version = pkgconfig('openbabel-2.0', '--modversion')
```
ghutchis: How about you try both openbabel-2.0 and openbabel?

scripts/python/setup.py
```diff
@@ -90,6 +91,16 @@ def make_release_tree(self, base_dir, files):
 
 class CustomBuildExt(build_ext):
     """Custom build_ext to set SWIG options and print a better error message."""
+    def find_swig(self):
```
ghutchis: This is fine. Probably worth finding out other names for SWIG binaries..

ghutchis (changes requested):

baoilleach:
There seem to be several changes here, unrelated to the problem the patch is intended to solve. Which makes me uneasy.

ghutchis:
I'm closing this - it's moot with modern SWIG and Python. Thanks

PR #1666, opened 2017-10-15
Minor fixes in the nwchem format reader

xomachine:
* Removed '@' from zts calculation patterns (it can be both of '@zts' or just '@' depending on nwchem version)
* Fixed non-matching amount of frequencies and intensities. (in particular in avogadro)
* Removed redundant vector<vector3>() insertion (it was already inserted above, and one more insertion per iteration just consumes memory.)


xomachine:
The change in nwchem that breaks zts reader: https://github.com/jeffhammond/nwchem/commit/76d2b8c465fd291883f3b8533ccda0686193bb8b

ghutchis:
Can you give me a bit more on this? I guess Jeff's change broke zts? Can we have a test file?

xomachine:
The example of the file prior the change can be found in nwchems QA tests. [ch4_zts.out](https://gist.github.com/xomachine/36f01755627c29829b86a3347c445513#file-ch4_zts-old-out-L405)([ch4_zts.out at the nwchem repository](https://raw.githubusercontent.com/jeffhammond/nwchem/master/QA/tests/ch4_zts/ch4_zts.out))
After the change, nwchem produces files like [this](https://gist.github.com/xomachine/bd1334f865ef519a26f3a55fc262cfb7#file-ch4_zts-new-out-L405).
To keep nwchem parser compatible with both of version I just removed '@' part of the zts calculation pattern.

Latest commits should solve another problem introduced with https://github.com/jeffhammond/nwchem/commit/76d2b8c465fd291883f3b8533ccda0686193bb8b . The equal sign now is truncated from  'Bead number = X' pattern after adding 'zts' to the start for the string. (Line 147440 in [new ch4_zts.out](https://gist.github.com/xomachine/bd1334f865ef519a26f3a55fc262cfb7) and 178025 in [old ch4_zts.out](https://gist.github.com/xomachine/36f01755627c29829b86a3347c445513))

I'm not sure how to write the test for that case. Including outputs mentioned above to the repository probably is not a good idea.

ghutchis:
Ok, fair enough.

PR #1670, opened 2017-10-20
Add support for wrapping GetRGB() call to return r, g, b params.

ghutchis:


baoilleach:
Sure, if the code works. I wonder though should I just return an int composed of ``R<<16 | G << 8 | B``. It's a bit of a faff whatever way it's done of course.

ghutchis:
I think this patch leads to a clean syntax in Python:
(red, green, blue) = ob.GetRGB(atom.atomicnum)

PR #1671, opened 2017-10-20
Create CODE_OF_CONDUCT.md

ghutchis:
Add a code of conduct - generally hasn't been a problem, but better to be up-front.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,53 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation
+in our project and our community a positive, harassment-free experience for everyone, regardless of age, body size, disability, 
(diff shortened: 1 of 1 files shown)


baoilleach:
Sure. I note though that it's not going to be confidential if sent to openbabel-devel@lists.sf.net, but the text gives that impression.

ghutchis:
Good point. I was thinking more along the lines of someone saying "hey, I have an issue" but I'd welcome either:
a) a way to provide confidentiality (since we don't have a separate maintainers list)
b) re-wording for the text

Not sure of either, although I can take a stab at rewording.

baoilleach:
Just make it clear that it's not confidential, but that they can email the lead developer confidentially if they want. I don't think you need to provide an email address - they'll figure it out.

baoilleach:
Where does this appear, BTW?

ghutchis:
It doesn't automatically appear anywhere. Here's an example for atom:

https://github.com/atom/atom
(it's linked from the README)
https://github.com/atom/atom/community

Issue #1615, opened 2017-08-15
pybel python 3 compatibility


matk86:
* error: `TypeError: a bytes-like object is required, not 'str'`
* in line https://github.com/openbabel/openbabel/blob/master/scripts/python/pybel.py#L528
* solution: before that line set `filename = filename.decode()`

cowsandmilk:
solution is more complex than that, you are describing an error that only occurs if the user passes in a bytes-like object as the filename. If, as I would expect to be typical, the user passes in a str, your solution would result in "AttributeError: 'str' object has no attribute 'decode"

matk86:
well, you could do a hasattr check then...

matk86:
`filename = filename.decode() if hasattr(filename, "decode") else filename`

baoilleach:
Looking at the docs for os.path, it seems that both str and byte filenames should be handled as much as possible. A possible solution is to think about path-like objects (https://docs.python.org/3/glossary.html#term-path-like-object).

mwojcikowski:
I assume that it will also break for Py2 and filename of unicode type. Maybe an input validation `isinstance(filname, str)` is sufficient? If not throw `ValueError` and let the user bother.

The other solution would be more complex and would require casting from bytes to str in Py3 and from unicode to bytes in Py2 with addition of `six` preferably.

afonari:
In what case filename is not a string? Just for my education.

baoilleach:
See top of https://docs.python.org/3/library/os.path.html.

afonari:
Holy sh*t! Thanks, @baoilleach.

PR #1673, opened 2017-10-24
use isinstance to test if filename is bytes

cowsandmilk:
Address #1615 by testing if filename is an instance of bytes and use a
bytes suffix in that case. This seems to be the approach used in cpython
source code, so I take it for being pythonic at the moment. (e.g.
https://github.com/python/cpython/blob/3f9183b/Lib/posixpath.py#L119 )

### DIFF
--- a/scripts/python/pybel.py
+++ b/scripts/python/pybel.py
@@ -250,8 +250,13 @@ class Outputfile(object):
         if not formatok:
             raise ValueError("%s is not a recognised Open Babel format" %
                              format)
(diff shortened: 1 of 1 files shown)


mwojcikowski:
I think, this will not work on Python 2. Id recommend using `six`:

```python
import six

isinstance(filename, six.string_types):

isinstance(filename, six.binary_type):
    pass
```

mwojcikowski:
On a second thought it might be ok, my bad. Although I always prefer the `six` way.

LGTM

PR #1677, opened 2017-10-26
Add support for %(NNN) notation for SMILES ring closures

baoilleach:
The % notation only supports up to 99 ring closures. I have an example of a SMILES string from RDKit that goes beyond this (included in the tests).

An extension to this is the %(...) notation which could in theory support any arbitrary number. I've implemented this with support for up to 5 digits. I'm also in the process of proposing this change to RDKit.

I can't see any downside to supporting this notation on reading. On writing, the only alternative is to refuse to write the SMILES.

I've also tightened up the parsing of the normal % notation. The existing code accepted non-digits.

(Apologies for the small number of unrelated white-space changes - my editor seems to be doing this without asking.)


merkys:
Is ``%(NNN)`` notation standard or is proposed for addition? As per OpenSMILES, it is possible to reuse already used bond IDs, what avoids the problem, unless one needs to have >99 open rings at time.

baoilleach:
It's not standard, it's an extension. The example testcase I provide is a real-life SMILES string that contains more than 99 open rings. We don't have control over what SMILES strings other tools generate.

merkys:
Regarding SMILES – I agree. Many software packages have their own micro-forks of SMILES. Hopefully they will not contradict.

For sure I would be more than glad to get our code together :)

baoilleach:
And just to be clear, RDKit doesn't (yet) have this extension. However, it does generate SMILES that require it. Both OB and RDKit will write %100 at the moment in this situation (if it ever occurs), which is incorrect.

merkys:
Indeed, ``%(XYZ)`` is better than just ``%XYZ``. The former would be recognized as syntax error by strict OpenSMILES parser whilst the latter would be broken up to closures of two rings, ``XY`` and ``Z``.

bbucior:
Tagging #1529

baoilleach:
See also rdkit/rdkit#1624

ghutchis:
I'm in favor. It seems like the OpenSMILES list has adopted the enhancement, and RDKit has also merged.

PR #1686, opened 2017-11-09
Clarify option needed to generate SWIG bindings.

jeffjanes:
If you try build bindings which have not been generated,
it provides a suggestion which, when copied into the cmake
command line, does not work.  -DRUN_SWIG is not a switch,
it requires an argument.


baoilleach:
Thanks

jeffjanes:
Sorry, did you intend to merge this rather than close it?  Or is there something I need to do to it first?

baoilleach:
Drat. My very bad... Will fix ASAP

PR #1687, opened 2017-11-09
Correct spelling of file name for Perl bindings

jeffjanes:
The MANIFEST file has openbabel-perl.cpp spelled with
an underscore rather than a dash.  This causes a
confusing but otherwise harmless warning when running
`perl Makefile.PL`.


ghutchis:
Oof, shows how much the Perl bindings are used - thanks! 👍

PR #1697, opened 2017-11-22
Expose obfunctions api through python bindings

cstein:
A recent API change (#1576) broke FragIt (https://github.com/FragIt/fragit-main) which depended on the knowledge about implicit bonds. Now, #1576 did include new utility functions (obfunctions.h) but these functions were never exposed through the python API.

1. This PR exposes those functions hidden away in obfunctions.h so I can use them in FragIt

This PR does not add or remove anything in the API apart from exposing functionality in the API through the python-bindings but I am unsure whether I am required to add some form of test or other documentation?

### DIFF
--- a/scripts/openbabel-python.i
+++ b/scripts/openbabel-python.i
@@ -31,6 +31,7 @@
 
 #include <openbabel/ring.h>
 #include <openbabel/obconversion.h>
+#include <openbabel/obfunctions.h>
 #include <openbabel/oberror.h>
 #include <openbabel/plugin.h>
 #include <openbabel/fingerprint.h>
@@ -261,6 +262,7 @@ namespace std { class stringbuf {}; }
 %include <openbabel/oberror.h>
 %include <openbabel/format.h>
 %include <openbabel/obconversion.h>
+%include <openbabel/obfunctions.h>
 %include <openbabel/residue.h>
 %include <openbabel/internalcoord.h>
 %include <openbabel/atom.h>


baoilleach:
Just a note that obfunctions.h is a placeholder for the final location of
these methods, which is something we need to think about.

Just for my own interest, how are you using these functions?

cstein:
Thanks for commenting @baoilleach :)

As long as the functions are exposed (or there is another clever way to easily calculate what I need) then I see no problem with where things are stored in the C++ API.

I used to use the `GetImplicitValence()` method on atoms in combination with `GetValence()` to figure out whether or not an atom needed to be given a hydrogen. Such a situation can arise if one needed to cut (in openbabel technically remove) a bond between a QM region and MM region with FragIt. I found an old blogpost of mine on the subject from almost 6 years ago (http://pythonchem.blogspot.dk/2012/01/adding-hydrogen-atoms-is-no-mean-feat.html) which explains it with a slightly different angle. The code I used has barely changed.

The updated code (see https://github.com/FragIt/fragit-main/blob/develop/src/qmmm.py#L139) needs to know how many bonds really would like to be present and the following code apparently satisfies my testcases.

Hope that explains how I used the old (and new) functionality.

baoilleach:
It does, but there may be a better approach if at the point of breaking the bond you know the exact number of hydrogens. The common idiom is to increase the hydrogen count by the bond order of the bond you are breaking. E.g. if you break H2C=CH2, then you should increase the implicit H count on both Cs by 2.

cstein:
Ok, that's perhaps an even more elegant solution - I assume the previous code (GetImplicitValence) sort-of did that.

I'll see if it makes sense to modify the code appropriately. The problem is now maintaining compatibility between that API change. I'll figure something out and of course I'll know if things broke either on your end or mine. Thanks a bunch for taking the time Noel.

baoilleach:
I'm afraid not. The GetImplicitValence() was more like GuessImplicitValence(), just like the functions in obfunctions.h. Most of the time it's fine of course, because most atoms are carbon.

Well, since I made the API breakage (for good reason - about 10 or 15 bugs were closed with this change), I should really be on hand to help sort out problems :-)

cstein:
:) and I appreciate the effort. Ok, I'll be on the mailing list when (and if) I need help with understanding your proposal about the implicit H count - but off the bat it sounds like it would be doable at some point.

PR #1698, opened 2017-11-22
Correct handling of atom maps

baoilleach:
To begin with, this PR addresses the following bugs:
1. Atom maps were not being updated when atoms/hydrogens were deleted
2. Atom maps were not being copied on adding two molecules together

It also changes the behaviour of the OBAtomClassData somewhat:
1. If an atom does not have any class, the value of 0 is returned (instead of -9999)
2. Setting a value of zero removes the value
3. It is not enforced, but the documentation now states that atom classes can only be positive integers in keeping with the Open Smiles spec (previously it was documented as allowing any integer)

While these changes make the library work as expected, I plan to submit an Enhancement Proposal to suggest that we either use AtomIds to store this data (instead of AtomIdxs), or store it directly on the OBAtoms (instead of on the OBMol).


ghutchis:
Personally, I'd vote towards storing it with the OBAtoms.

PR #1699, opened 2017-11-23
Speed up the SMILES writer

baoilleach:
This PR makes SMILES writing fast.

The main speedup was to avoid regenerating an OBStereoFacade each time we needed to look up a single atom. This was occurring in multiple places. Now a single OBStereoFacade is used for the molecule.

Also, I have removed all uses of strcpy, strlen and strcat.

There is some room for further speedups but I might move onto other stuff for now.

### DIFF
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -2254,14 +2254,12 @@ namespace OpenBabel {
                                 vector<unsigned int> &canonical_order,
                                 OBCanSmiNode *node);
     void         CreateFragCansmiString(OBMol&, OBBitVec&, bool isomeric, bool kekulesmi, std::string&);
-    bool         GetTetrahedralStereo(OBCanSmiNode*,
+    const char * GetTetrahedralStereo(OBCanSmiNode*,
                                       vector<OBAtom*>&chiral_neighbors,
-                                      vector<unsigned int> &symmetry_classes,
-                                      char*);
-    bool         GetSquarePlanarStereo(OBCanSmiNode*,
+                                      vector<unsigned int> &symmetry_classes);
+    const char*  GetSquarePlanarStereo(OBCanSmiNode*,
                                        vector<OBAtom*>&chiral_neighbors,
-                                       vector<unsigned int> &symmetry_classes,
-                                       char*);
+                                       vector<unsigned int> &symmetry_classes);
     bool         GetSmilesElement(OBCanSmiNode*,
                                   vector<OBAtom*>&chiral_neighbors,
                                   vector<unsigned int> &symmetry_classes,
@@ -2576,7 +2574,7 @@ namespace OpenBabel {
   {
     char symbol[10];
     symbol[0] = '\0'; // make sure to initialize for all paths below
-    char bracketBuffer[32];
+
     bool bracketElement = false;
     bool normalValence = true;
     bool writeExplicitHydrogen = false;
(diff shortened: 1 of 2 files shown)


ghutchis:
Have you been doing any performance testing ([g]perf, profiling, sampling, etc.)

It would be interesting to see where the hot-spots are in the unit tests.

baoilleach:
I used a profiler to find the stereofacade issue. That was essentially a misuse of the class. After that, there are a few more things that could probably be rewritten for speed such as CreateCanonTree, testing whether two stereo configs are equal, and finding the smallest ring a bond is in (maybe this latter is unnecessary in the smiles code).

More generally, calling buffer.reserve at the start might help, caching option lookups, caching multiple accesses to GetIdx (etc).

ghutchis:
I'm wondering more whether the "hot spots" are here or elsewhere.

baoilleach:
I will be writing up a blog post and will include further details.

baoilleach:
Well I started writing a blog post, but when I looked into it closely, I thought I'd better trim a few of more obvious ones:

> There's an intial slow period where OB is loading the plugins (or at least, that's my interpretation). Focusing on the plateau, it seems that 63% of the time is spent writing the molecule, while 31% is spent reading. This is both good and bad; I would have expected reading to be slower than writing, due to the associated mallocing; on the other hand, it may indicate that there is potential for optimising.
>
> Stepping down through the functions, there are some obvious small gains. 1.6% of the total time is spent copying the description of the format so that the first portion can be used in an audit message. The fix here is probably to get rid of audit messages altogether, although avoiding the copy would an alternative. 3.4% of the time is spent on writing std::endl to the output buffer. "\n" might be a sensible alternative, though one could imagine having the option if you really needed it.
> 
> In the main part of the writer (52.3% of total), 10.2% is spent on CreateCisTrans, while 38.2% is spent on the meat of the writing. That 10.2% is suspicious; when you look into it, almost all the time is spent on FindSmallestRing, which in turn calls GetSSSR(). Rumour has it that SSSR is the root of all evil, and so I'm pretty sure this can be avoided.
> 
> As regards the 38.2%, ToCansmilesString() is 14.4%, BuildCanonTree() is 8.8% and operator << is 10.5% (!). That last one puzzled me, but then I found it. This is the generation of a string containing the canonical atom order. Wow, stringstream is slow. I think we can do something about this, like avoid it altogether unless requested (in which case, avoid stringstream).
> 
> Another optimisation in multiple places would be to cache the values of various options to avoid multiple lookups of the option map (2.3% in total).
> 
>

ghutchis:
I had some plans for eliminating GetSSSR - can you find me some of the call-tree for those? What is calling FindSmallestRing?

There's a fairly simple strategy to find ring closures - you do a bounded breadth-first search from a particular atom. If you haven't found your way back to the atom after X bonds, then you're probably not in a ring. The catch is for big macrocycles, but after discussion with a few people:
- X = 14 should be sufficient for real aromatics
- X = 20 should be more than sufficient for most cases

baoilleach:
For sure, and actually X just needs to be 8 in this case:
```
#define IMPLICIT_CIS_RING_SIZE 8

        // Do not output cis/trans bond symbols for double bonds in rings of size IMPLICIT_CIS_RING_SIZE or less
        OBRing* ring = dbl_bond->FindSmallestRing();
        if (!ring || ring->Size()>IMPLICIT_CIS_RING_SIZE)
          _cistrans.push_back(*ct);
      }
```
Hmmm...I hope there's a check to see whether the bond is in a ring before calling the SSSR. Also, but this needs more thought, I don't think the writer is the correct place for this code. This could be another case of defaulting to accepting it directly when reading SMILES, but I have to think some more about this. For sure, the more straightforward fix right now would be to do a bounded BFS.

baoilleach:
Just a note, I've started work on the BFS...

PR #1706, opened 2017-11-26
Faster SMILES: Replace std::endl by "\n"

baoilleach:
3.5% of the time for SMI->SMI conversion is spent on writing std::endl, which flushes the buffers. I've replaced it with just "\n" for non-Windows, and "\r\n" for Windows. Does this need a further 'ifdef' for the Mac @ghutchis?

I've also edited the code to have just a single write to the output file stream. I note in passing that GetTitle(), by default, seems to be doing more work than necessary (a job for another day).


ghutchis:
Mac OS X uses Unix line endings, since ~2001. I think we're safe.

PR #1707, opened 2017-11-26
Remove some audit messages

baoilleach:
Remove some audit messages that copy the entire format description and then substring it.

For smi->smi conversion, 1.6% of the total conversion time was spent on one of these audit message blocks, the majority of which on the string copy of the format description (which is not small for SMILES).

Since the other ones are similar, why not remove all three? Actually, I would remove all audit and debug level logging statements throughout the library (I have never used them for debugging), so I have no qualms about removing these three. They could be rewritten to avoid the string copy, but... What do they think?

### DIFF
--- a/src/obmolecformat.cpp
+++ b/src/obmolecformat.cpp
@@ -37,13 +37,6 @@ namespace OpenBabel
 
     OBMol* pmol = new OBMol;
 
-    std::string auditMsg = "OpenBabel::Read molecule ";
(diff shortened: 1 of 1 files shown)


ghutchis:
I'm totally fine removing these. I don't want to remove audit-level statements, because I think it's useful to retain a list of anything that possible changes the state of the molecular data.

Debugging-level.. that's a different story.

PR #1708, opened 2017-11-27
Fix a memory leak to due OBGraphSym having a virtual destructor

baoilleach:
Dr Memory flagged this up as a memory leak. I can't remember if it's just when used in the Python bindings or not. Either way, only abstract base classes should have a virtual destructor.


baoilleach:
Actually - let's put this on hold. Let me confirm this for sure....

PR #1710, opened 2017-11-27
Speed up the SMILES writer by avoiding creating the SMILES atom order vector

baoilleach:
Creating the SMILES atom order OBPairData was taking 10% of the time of SMI->SMI conversion. I rewrote it to avoid using stringstream and I made it optional.


baoilleach:
Uh oh - there's a whole lot of whitespace changes in there that I didn't intend. I'll cancel this and try again another day....

PR #1712, opened 2017-11-28
Make generation of the SMILES atom order vector optional (Take 2)

baoilleach:
Creating the SMILES atom order OBPairData was taking 10% of the time of SMI->SMI conversion. I rewrote it to avoid using stringstream and I made it optional.

### DIFF
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -145,6 +145,8 @@ namespace OpenBabel {
         "     This gives canonical labels 1,2,3,4 to atoms 4,2,1,3 respectively,\n"
         "     so that atom 4 will be visited first and the remaining atoms\n"
         "     visited in a depth-first manner following the lowest canonical labels.\n"
+        "  O  Store the SMILES atom order as a space-separated string in an\n"
+        "     OBPairData with the name 'SMILES Atom Order'\n"
         "  F  <atom numbers> Generate SMILES for a fragment\n"
         "     The atom numbers should be specified like \"1 2 4 7\".\n"
         "  R  Do not reuse bond closure symbols\n"
@@ -2222,7 +2224,6 @@ namespace OpenBabel {
     OBBitVec _uatoms,_ubonds;
     std::vector<OBBondClosureInfo> _vopen;
     unsigned int _bcdigit; // Unused unless option "R" is specified
-    std::string       _canorder;
     std::vector<OBCisTransStereo> _cistrans, _unvisited_cistrans;
     std::map<OBBond *, bool> _isup;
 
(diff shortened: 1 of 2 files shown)


mwojcikowski:
While you're at it, is it possible to expose this function to a python API? I think most of us would like to use it to get a list.

baoilleach:
You mean, instead of the string? Data can only be returned via
obgenericdata.

ghutchis:
I'm totally fine avoiding the stringstream. I'm not so sure about making it optional.
- Regular SMILES - is the canonical atom order useful? The atoms won't be in that order
- Canonical SMILES - is the overhead that high now? As @mwojcikowski indicates, it would be useful for many reasons.

baoilleach:
Just to clarify first, the string is the SMILES output order rather than the canonical atom order. Even where it is a regular SMILES, the output order may not be the order of the atoms in the molecule.

I don't see the downside to making it optional. It avoids (a) the string generation, (b) the original vector generation (though this PR doesn't address that), (c) a minor slowdown in GetData() due to the additional entry, (d) the memory footprint of the OBPairData(). Sure, it might not be much, but why do it at all? Once you start looking at larger molecules, these are going to add up and I would expect that they will affect the size of the biggest molecule we can handle.

baoilleach:
@mwojcikowski Given that this is SMILES output order, I don't think it would make sense as a method of a Molecule.

mwojcikowski:
Is there some other way to get some kind of "canonical" atoms order? At least that's what I use it for. 

On the other hand, there was a discussion on RDKit's mailing list that their function of ranking atoms do not match the one in smiles. So some might even consider this a "feature".

baoilleach:
The canonical labels are already available through the API, via OpenBabel::CanonicalLabels. See also OBGraphSym.GetSymmetry to get the symmetry classes.

ghutchis:
I don't think discussion of getting the canonical labels needs to hold up this merge.

baoilleach:
Just FYI, here is a Python example of getting the canonical labels:
```
>>> import pybel
>>> ob = pybel.ob
>>> mol = pybel.readstring("smi", "c1(F)ccc(F)cc1").OBMol
>>> gs = ob.OBGraphSym(mol)
>>> symmetry = ob.vectorUnsignedInt()
>>> gs.GetSymmetry(symmetry)
3
>>> list(symmetry)
[3, 1, 2, 2, 3, 1, 2, 2]
>>> canonical_labels = ob.vectorUnsignedInt()
>>> ob.CanonicalLabels(mol, symmetry, canonical_labels)
>>> list(canonical_labels)
[2, 1, 3, 5, 7, 8, 6, 4]
```

PR #1714, opened 2017-12-02
Avoid logging messages that are taking time

baoilleach:
ThrowError was taking a total of 1.9% of the time for smi->smi. This was mainly attributed to three log messages, all audits. I've added an 'if' statement to check the logger's output level before throwing the error. I think it would make sense for this to be an idiom throughout the library.

### DIFF
--- a/src/mol.cpp
+++ b/src/mol.cpp
@@ -1433,8 +1433,9 @@ namespace OpenBabel
 
   bool OBMol::Clear()
   {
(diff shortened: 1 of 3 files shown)


ghutchis:
Yes, it definitely makes sense as an idiom.

baoilleach:
Actually, if that's the case, we can bring back the log statements I removed in an earlier PR, and just put them in an 'if'. Just say the word.

PR #1715, opened 2017-12-04
Speed up SMILES writer by replacement of SSSR in SMILES writer with a bounded BFS

baoilleach:
To test whether a double bond is implicitly cis (and thus requires no stereobond symbols), the original code used SSSR based ring perception. In fact, it always triggered it, even where the bond was not in a ring. Doh!

The replacement code does not use the SSSR.

### DIFF
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -2378,7 +2378,7 @@ namespace OpenBabel {
           continue;
         // Do not output cis/trans bond symbols for double bonds in rings of size IMPLICIT_CIS_RING_SIZE or less
         unsigned int boundedringsize = OBBondGetSmallestRingSize(dbl_bond, IMPLICIT_CIS_RING_SIZE);
-        if (boundedringsize > 0)
+        if (boundedringsize == 0) // either not in ring at all, or not in small ring
           _cistrans.push_back(*ct);
       }
     }
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -2364,6 +2364,57 @@ namespace OpenBabel {
     return(idx);
   }
 
+  // Is this dbl bond in a ring that is of size <= IMPLICIT_CIS_RING_SIZE?
+  //
+  // A bounded BFS from one side of a bond trying to find the other side.
+  //   - The required queue is implemented using a std::vector for efficiency
+  //   - Note that items are never popped from the queue, I just move a pointer
+  //     along to mark the left hand side
+  //   - A potential improvement would be to BFS from both sides at the same time
+  static bool IsNotInRingOfSizeThatImpliesCis(OBMol *mol, OBBond *bond)
+  {
+    if (!bond->IsInRing()) return true;
+    OBAtom* start = bond->GetBeginAtom();
+    OBAtom* end = bond->GetEndAtom();
+    std::vector<OBAtom*> qatoms;
+    unsigned int numatoms = mol->NumAtoms();
+    qatoms.reserve(numatoms < 42 ? numatoms : 42); // Value chosen by inspection on ChEMBL when using maxsize of 8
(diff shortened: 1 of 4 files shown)


src/formats/smilesformat.cpp
```diff
@@ -2364,6 +2364,57 @@ namespace OpenBabel {
...
   }
 
+  // Is this dbl bond in a ring that is of size <= IMPLICIT_CIS_RING_SIZE?
+  //
+  // A bounded BFS from one side of a bond trying to find the other side.
+  //   - The required queue is implemented using a std::vector for efficiency
+  //   - Note that items are never popped from the queue, I just move a pointer
+  //     along to mark the left hand side
+  //   - A potential improvement would be to BFS from both sides at the same time
+  static bool IsNotInRingOfSizeThatImpliesCis(OBMol *mol, OBBond *bond)
```
ghutchis: I'd actually prefer if we can make this a general method, e.g., OBMolAtomBFSIter. Many of the calls to SSSR could be replaced by an iterator like this (e.g., aromaticity)

src/formats/smilesformat.cpp
```diff
@@ -2364,6 +2364,57 @@ namespace OpenBabel {
...
+      OBAtom* curr = qatoms[qstart];
+      if (qstart == depthmarker) {
+        depth++;
+        depthmarker = qatoms.size();
+      }
+      qstart++;
+      if (seen.BitIsSet(curr->GetIdx()))
+        continue;
+      seen.SetBitOn(curr->GetIdx());
+      if (depth < IMPLICIT_CIS_RING_SIZE) {
```
ghutchis: IMPLICIT_CIS_RING_SIZE should be a method parameter

test/testbindings.py
```diff
@@ -88,6 +88,21 @@ def testAtomMapsAfterDeletion(self):
         mol.removeh()
         self.assertEqual(mol.write("smi", opt={"a":True}).rstrip(), "C[NH:2]")
 
+    def testImplicitCisDblBond(self):
```
ghutchis: This came up in the coordinate generation discussions... what should the code do when presented with a SMILES with a trans indication in a small ring? Right now, we don't have any sort of stereo warning.

baoilleach: Actually I almost did this. Well I did it, and then reverted, to be honest. But I can put it back. My only concern is that while this is efficient for a single atom, I think a different approach would be better if you are interested in identifying all atoms in smallish rings (I had a discussion on this with @johnmay).

But for sure, there's no harm in making the method available.

src/formats/smilesformat.cpp
baoilleach: Actually I almost did this. Well I did it, and then reverted, to be honest. But I can put it back. My only concern is that while this is efficient for a single atom, I think a different approach would be better if you are interested in identifying all atoms in smallish rings (I had a discussion on this with @johnmay).

But for sure, there's no harm in making the method available.

src/formats/smilesformat.cpp
baoilleach: Sure.

test/testbindings.py
baoilleach: Sure. It could be part of the PerceiveStereo.

baoilleach:
I think these changes address your comments (apart from the comment about trans stereo in small rings).

baoilleach:
Or did you have in mind an actual iterator?

ghutchis:
I had a bounded iterator in mind, but I can take it from there. No reason to hold up this patch, except for the weird Travis fail.

baoilleach:
Hmmm...seems to work fine on Windows (Travis doesn't run the Windows tests, but I've just done them here). I'll check it out on Linux...

baoilleach:
After a reboot, the Windows tests did fail - this does happen occassionally where Windows refuses to believe the dlls have changed. Anyway, dumb mistake - now fixed.

PR #1718, opened 2017-12-06
Minor speedup to avoid repeated calls to IsOption by caching some options

baoilleach:
In the SMILES writer, the profiler showed IsOption() as talking up a percent or two. I've cached a few of the options in the functions highlighted by the profiler. Is it possible to persist some variables between repeated calls to the writer (or would this cause some threading issues) as this would be a potential improvement?

### DIFF
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -2218,9 +2218,12 @@ namespace OpenBabel {
     bool kekulesmi;
     bool showexplicitH;
     bool smarts;
-    OutOptions(bool _isomeric, bool _kekulesmi, bool _showexplicitH, bool _smarts):
+    const char* ordering; // This is a pointer to the string in the original map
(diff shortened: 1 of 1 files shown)


ghutchis:
It would depend on how the variables are retained, as far as threading goes.

baoilleach:
Just reran the timings. Locally, based on master + AvoidSSSR, IsOption takes 3.8%. With this change, it's 2.0%, with 0.6% from BuildCanonTree, 0.3% from WriteMolecule, and 0.3% from CreateCansmiString.

One thing to think about is whether we could speed up IsOption itself, e.g. think about what sort of std::map we use, or whether to use a map at all for single letter boolean options.

If you have no issues, I think for the moment that I'm happy with this PR as it is, and leave the more complicated issue of retaining state for another day.

baoilleach:
Actually - maybe there's one or two more options that could be cached. Could you leave this here for the moment, and I might push another commit later this week...

baoilleach:
Ok - ready to go. Down to 1.3%.

PR #1720, opened 2017-12-07
Read reaction map from ChemDraw CDX files

CamAnNguyen:
[test-reaction-map.zip](https://github.com/openbabel/openbabel/files/1538600/test-reaction-map.zip)

This PR attempt to fix the following bug:

OpenBabel get Segfault while reading CDX contains such a reaction map like this: A -> B -> C (I expect to get the 2 reactions from the output: A->B and B->C). I include a simple CDX to test for this case (please rename it from `.zip` to `.cdx`).

### DIFF
--- a/src/formats/chemdrawcdx.cpp
+++ b/src/formats/chemdrawcdx.cpp
@@ -182,6 +182,8 @@ private:
   std::map<CDXObjectID, graphicType> _graphicmap;
   std::map<CDXObjectID, OBMol*> _molmap;
   std::map<CDXObjectID, std::vector<CDXObjectID> > _groupmap;
+  // In case of chain A -> B -> C, B is both reactant and product
+  CDXObjectID _lastProdId;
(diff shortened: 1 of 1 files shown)


baoilleach:
It would be interesting to hear more about why the segfault occurs in the first place. The shared_ptrs were added to facilitate exactly this use case; multiple reactions using the same molecules.

baoilleach:
...but the change looks fine. Thanks.

CamAnNguyen:
Hello, sorry that I missed your comment somehow. 

From [here](https://github.com/openbabel/openbabel/blob/master/src/formats/chemdrawcdx.cpp#L202):  `When the reaction is complete it is output via AddChemObject()`.
So the reaction is output within during `pInFormat->ReadChemObject(this);`(from [`obconversion.cpp`](https://github.com/openbabel/openbabel/blob/master/src/obconversion.cpp#L545)). 

Personally, I prefer to fully `read` first then `output` the data. But I do not want to change too much since I am quite new to the codebase. That leads to the addition memory allocating.

PR #1721, opened 2017-12-07
Fix uninitialized read in kekulize.cpp found by Dr Memory.

baoilleach:


ghutchis:
This looks like a false positive to me, but it's harmless. I don't see how, if atom->GetIdx() is uninitialized, assigning it to a variable is going to initialize it.

As I said, it's harmless, so assuming the builds pass, I'm fine with merging.

PR #1724, opened 2017-12-08
Make sure that OBAtom::IsInRing always triggers ring perception if not set as perceived

baoilleach:
Previously, if the OBAtom was marked as a ring atom, then no ring perception was done, even where the rings were marked as unperceived. It should be the case that when a perceived property is marked as unset, that perception will always be triggered when queried.

I made a similar change recently for IsAromatic(), and it looks like there are more SSSR ring-related functions that may have the same problem.

### DIFF
--- a/src/atom.cpp
+++ b/src/atom.cpp
@@ -801,18 +801,14 @@ namespace OpenBabel
 
   bool OBAtom::IsInRing() const
   {
-    if (((OBAtom*)this)->HasFlag(OB_RING_ATOM))
-      return(true);
-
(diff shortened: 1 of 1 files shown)


baoilleach:
Here is the previous commit I reference:
https://github.com/openbabel/openbabel/pull/1696/commits/63173f51907d6e7d4fe866708680065dcca9b624

ghutchis:
I'm merging this, but if you have time to look for similar functions, I'd appreciate it.

PR #1725, opened 2017-12-13
Two minor SMILES speed improvements

baoilleach:
Adding a few std::vector::reserve() made a measurable difference. In the profiler (with VS2017), this is attributed to _emplace_back.

Rewriting the parsing loop to be entirely switch-statement-based should be more efficient, but I didn't notice any speed-up.


ghutchis:
Can you submit a PR with the first change? I'm happy to merge that, but it seems the second fails Travis.

baoilleach:
Hmmm....another one that seems to only fail on Linux.

baoilleach:
I've re-pushed just the initial commit on its own. I haven't tracked down the other problem yet but will deal with it on a different PR.

PR #1733, opened 2017-12-26
warning/error messages for fastindex when the structure file is compressed

adalke:
obabel has no problem creating a .fs file from a .sdf.gz file.

However, the search will fail because the .fs file stored the record index in the uncompressed file, while the datastream.seekg() seeks to a position in the compressed file, and the code returns compressed text.

As a result, the fastsearch results are essentially useless.

What this patch does is:
  1. print a warning message during index creation if the input file is gzip compressed
  2. log an error message and stop during search if the datafile is gzip compressed


ghutchis:
Makes sense, thanks!

PR #1737, opened 2017-12-28
Fix Python3 compatibility

ghutchis:


baoilleach:
I'd just change it to items() rather than introduce 'six'.

baoilleach:
...because six is not included with Python by default, AFAIK.

ghutchis:
Okay, I'll wait for the build to finish and then edit the commit message.

PR #1600, opened 2017-07-12
Gaussformat multiple charge modles patch

mmghahremanpour:
Here is implemented to read more than one charge models from the Gaussian output and store them as OBGenericData in OBMol. It now reads Mulliken, Hirshfeld, CM5, and ESP charges.

PR #1740, opened 2017-12-28
Store multiple partial charge models

ghutchis:
Cleanup of #1600 for discussing how to store multiple atomic partial charges

### DIFF
--- a/include/openbabel/generic.h
+++ b/include/openbabel/generic.h
@@ -1185,9 +1185,71 @@ namespace OpenBabel
 
   };
 
- //! A standard iterator over vectors of OBGenericData (e.g., inherited from OBBase)
-  typedef std::vector<OBGenericData*>::iterator OBDataIterator;
+  class OBAPI OBPcharge
+  {
+  protected:
+    int         _atomIndex;
+    double      _q;
+  public:
+    OBPcharge() {};
+    OBPcharge(int atomIndex, double q) {_atomIndex = atomIndex; _q = q ;}
+    ~OBPcharge() {};
 
+    double GetQ() {return _q;}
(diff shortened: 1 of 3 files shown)


ghutchis:
Paging @mmghahremanpour

include/openbabel/generic.h
```diff
@@ -1185,9 +1185,71 @@ namespace OpenBabel
...
+    double GetQ() {return _q;}
+    int    GetAtomIndex() {return _atomIndex;}
+    void   SetQ(double q) {_q = q;}
+    void   SetAtomIndex(int atomIndex) {_atomIndex = atomIndex;}
+    
+  };
+
+  typedef std::vector<OBPcharge*>::iterator OBPchargeIterator;
+  
+  class OBAPI OBPcharges: public OBGenericData
```
ghutchis: This makes little sense. I'm open to storing multiple sets of named partial charges from a QM file. No problem.

But this is hugely inefficient. For one, there's already support for an atomic partial charge OBAtom::SetPartialCharge() - which doesn't require extra storage.

Even if you want to have this class as a generic data type, the storage here is inefficient. Just keep a vector<double> for the charges? Why bother with the atom index business? Just index into the vector!

src/generic.cpp
```diff
@@ -1643,6 +1643,23 @@ unsigned int OBVibrationData::GetNumberOfFrequencies() const
   return this->_vFrequencies.size();
 }
 
+bool OBPcharges::FindPcharge(int    atomIndex,
+                             double *q)
```
ghutchis: This is hugely slow. It's completely un-needed if the charges are kept as vector<double> in the OBPCharges class.

Issue #1731, opened 2017-12-23
CML output error if atom class data exists AND more than 26 atoms have the same class


BernardUPSud:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: Open Babel 2.4.1 -- Jan  2 2017 -- 15:21:43
Operating system and version: Linux styx 4.9.0-4-amd64 #1 SMP Debian 4.9.65-3 (2017-12-03) x86_64 GNU/Linux

## Expected Behavior

First of all, thanks for the great software!

I use smile code to define AtomClass value for each atom and want to write it in CML output. For exemple `[N:730]`, should give the following output for the atom field:

`<atom id="aa730" elementType="N"/>`

The atom id is created including a string (aa) and the atom class value (730). The string value is incremented (ab, ac, ...) if other atoms with the same atom class exist in the molecule.

## Actual Behavior

Everything is fine if and only if the number of atoms with the same class is smaller than 27. For large molecules, obabel fails and a seg fault is produced.

When using obabel 2.4.1 with the command
`obabel -ismi mpdtmc3.smi -ocml`, a segmentation fault occurs with the original version whereas a unique atom id is given for each atom with the correction proposed.

[mpdtmc3.smi.zip](https://github.com/openbabel/openbabel/files/1584141/mpdtmc3.smi.zip)
[cmlformat.cpp.zip](https://github.com/openbabel/openbabel/files/1584138/cmlformat.cpp.zip)

baoilleach:
I was just looking at this code this week, and I can imagine the problem. Can I ask what do you use this CML for? What I've been thinking about is whether it would be better to write out the atomclass information in some other way if this is allowed by the CML spec.

BernardUPSud:
Hi,

I am doing molecular simulations with empirical force fields (e.g. OPLSAA or other). In order to do a molecular simulation, I must assign a force field type to each atom in a molecule. Recently, I decided to use the AtomClass number as a force field type. I found that CML is the only output format that use the AtomClass number but I would like to output this information directly in tinker xyz file (instead of MM2 force field).

My work flow for a molecular simulation is then: create a SMILE code for molecules including AtomClass number (or equivalently force field type), use obabel to output a tinker xyz file (with coordinates and connectivity informations) and a cml file (with force field information) from which as script creates input files for my molecular dynamics code…

[...]

PR #1741, opened 2017-12-29
Refactor atom class to being data on an atom rather than on a molecule

baoilleach:
I've taken a chance on this one, rather than writing up the Enhancement Proposal. Here goes...

Atom class data is currently stored as an OBAtomClassData object on an OBMol. This PR moves it to being stored on an OBAtom directly.

The overall rationale for this change is not so much the conceptual issue of atom data being stored on the molecule object - sure, this is true, but it's something I could live with. The problem is that the data is currently stored using atom indices, and these require bookkeeping/updating when the molecule is modified. All of this bookkeeping goes away if the information is stored directly on an atom. This is a performance issue for a molecule with multiple atom classes - just consider that a call to DeleteAtom() ~~for an atom with an atom class~~ means that all of the existing atom classes may need to be adjusted.

Note that this doesn't yet address #1731, but I will do this next.


ghutchis:
It looks like there's a problem with strcmp in base.cpp:
https://stackoverflow.com/a/9404274/131896

src/formats/tinkerformat.cpp
```diff
@@ -152,8 +153,6 @@ namespace OpenBabel
     bool mmffTypes = pConv->IsOption("m",OBConversion::OUTOPTIONS) != NULL;
     bool mm3Types = pConv->IsOption("3",OBConversion::OUTOPTIONS) != NULL;
     bool classTypes = pConv->IsOption("c", OBConversion::OUTOPTIONS) != NULL;
-    if (pmol->GetData("Atom Class") == NULL)
-      classTypes = false;
```
ghutchis: Hmm. The point of this test was to switch to a different atom type (MM2) if atom classes aren't available. Not quite sure how to do that easily now (except looping through the atoms).

ghutchis: Not that this should hold up the change, though.

baoilleach: With the caveat that I'm not familiar with the format, I can see how the behavior is different, but I  prefer this behavior. If you leave out atom classes but specify 'c', you want to know that there's something not right. I could add a warning, but I'm not keen on an automatic fallback.

PR #1742, opened 2018-01-02
Reading different types of atomic partial charges from Gaussian log file

mmghahremanpour:
Cleaning up the code to read different atomic partial charges from Gaussian log file, namely Mulliken, Hirshfeld, CM5, and ESP charges.

### DIFF
--- a/src/formats/gaussformat.cpp
+++ b/src/formats/gaussformat.cpp
@@ -697,10 +697,24 @@ namespace OpenBabel
               if (!ifs.getline(buffer,BUFF_SIZE)) break;
             }
         else if(strstr(buffer,"Total atomic charges") != NULL ||
-                strstr(buffer,"Mulliken atomic charges") != NULL)
+                strstr(buffer,"Mulliken atomic charges") != NULL ||
+                strstr(buffer,"Mulliken charges:") != NULL)
           {
             hasPartialCharges = true;
             chargeModel = "Mulliken";
+            /*
+              Gaussian usually calculates the electronic 
+              properties more than once, before and after 
+              geometry optimization. The second one is what
+              we should be interested in. Thus, here, we 
+              delete the previously added Data to store the
+              new one.
+             */
+            if (mol.HasData("Mulliken charges"))
+              {
+                mol.DeleteData("Mulliken charges");
+              }
+            OBPcharge *Mulliken = new OpenBabel::OBPcharge();
             ifs.getline(buffer,BUFF_SIZE);	// column headings
             ifs.getline(buffer,BUFF_SIZE);
             tokenize(vs,buffer);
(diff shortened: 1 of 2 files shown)


mmghahremanpour:
Would you please review this patch?

src/formats/gaussformat.cpp
```diff
@@ -711,10 +725,64 @@ namespace OpenBabel
...
+            ifs.getline(buffer,BUFF_SIZE);	// column headings
+            ifs.getline(buffer,BUFF_SIZE);
+            tokenize(vs,buffer);
+            while (vs.size() >= 8 &&
+                   strstr(buffer,"Tot ") == NULL)
+              {
+                atom = mol.GetAtom(atoi(vs[0].c_str()));
+                if (!atom)
+                  break;
+                atom->SetPartialCharge(atof(vs[2].c_str()));
```
ghutchis: Wait - what's the preferred partial charge? I think there either needs a way to set the preferred charge (e.g. that will be written to .mol2 and other formats) or some discussion about what charges a user will get if multiple charges are found.

Right now, the code seems to quasi-randomly pick the atom->SetPartialCharge() based on the last set that it finds in the file.

Personally, it seems like I'd prefer something like:
- ESP if available
- Hirshfeld if not ESP
- Mulliken if nothing else

But it would be helpful to have a INOPTION to pick which type of partial charge is retained.

src/formats/gaussformat.cpp
```diff
@@ -711,10 +725,64 @@ namespace OpenBabel
...
+                                    
+              }
+            if (CM5->NumPartialCharges() != mol.NumAtoms() || 
+                Hirshfeld->NumPartialCharges() != mol.NumAtoms()) 
+            {
+                cerr << "The number of Hirshfeld or CM5 partial charges is not equal to the number of atoms!" << endl;
+                exit (-1);
+            }
+            Hirshfeld->SetAttribute("Hirshfeld charges");
+            CM5->SetAttribute("CM5 charges");
```
ghutchis: These should all have:
CM5->SetOrigin(fileformatInput);

(as in, they're not perceived Gasteiger or other charges)

src/formats/gaussformat.cpp
mmghahremanpour: This is actually needed. It would be helpful to have a INOPTION to pick which type of charge. We also need to consider the situation if the user specifies a charge type that does not exist. But, I think it is better to add these features as a new patch. 
Do you agree to have it as new patch?

ghutchis: That's fine.

mmghahremanpour:
> On Jan 11, 2018, at 8:48 PM, Geoff Hutchison <notifications@github.com> wrote:
> 
> @ghutchis commented on this pull request.
> 
> In src/formats/gaussformat.cpp <https://github.com/openbabel/openbabel/pull/1742#discussion_r161057678>:
> 
> >                }
> +            OBPcharge *Hirshfeld = new OpenBabel::OBPcharge();
> +            OBPcharge *CM5       = new OpenBabel::OBPcharge();
> +            ifs.getline(buffer,BUFF_SIZE);	// column headings
> +            ifs.getline(buffer,BUFF_SIZE);
> +            tokenize(vs,buffer);
> +            while (vs.size() >= 8 &&
> +                   strstr(buffer,"Tot ") == NULL)
> +              {
> +                atom = mol.GetAtom(atoi(vs[0].c_str()));
> +                if (!atom)
> +                  break;
> +                atom->SetPartialCharge(atof(vs[2].c_str()));
> That's fine.
> 
> 


I added SetOrigin() to the all charge types and now the updated one is in #1752.

ghutchis:
This superseded by #1752 so I'm closing this one.

PR #1743, opened 2018-01-02
Fix ChemDraw CDX incremental value

CamAnNguyen:
I am not sure if I am correct. But while implementing some more feature with ChemDraw CDX files, I found this one. 
The current value of `kCDXObj_Border` is `0x801a` which is the next value from `0x8019`, it should be `0x8020`.

### DIFF
--- a/data/chemdrawcdx.h
+++ b/data/chemdrawcdx.h
@@ -373,7 +373,7 @@ enum CDXDatumID {
 	kCDXObj_BracketedGroup,				// 0x8017
 	kCDXObj_BracketAttachment,			// 0x8018
 	kCDXObj_CrossingBond,				// 0x8019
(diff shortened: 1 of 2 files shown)


baoilleach:
I'd like to merge this, but when you say "I am not sure if I am correct", I don't know what to do. Is there some way you can decide whether you are correct? Or some way we can help.

CamAnNguyen:
I created a CDX file from ChemDraw and tested it. It could not get what I want with `border` attribute and this fix my issue (with my current version of ChemDraw). 
Technically, I am pretty sure about this "bug". Thing makes me confused is that this file is the header from ChemDraw themself so it just a little bit weird.

baoilleach:
That's sounds fine. I know there are some errors in that header file - I guess they don't use it themselves.

PR #1744, opened 2018-01-04
Ensure portability to ARM platforms

### LINKED ISSUE #1689: Compiliation fails on ARM platforms (invalid type conversion)
Compilation for ARM platforms using gcc 7.2 fails with `invalid conversion from 'signed char*' to 'const char*'`. I believe this is because on ARM the char data type is unsigned by default, but in `src/formats/pngformat.cpp` and `src/formats/yasaraformat.cpp` negative values are assigned. I used Open Babel version 2.4.1 but it’s still the same in the latest source files.

The following patch which replaces the negative integer with its two’s complement works, but I have no idea if that breaks something else. The best solution would probably be to use `singed char`.

```
--- src/formats/pngformat.cpp.orig      2016-09-21 21:55:37.000000000 +0200
+++ src/formats/pngformat.cpp   2017-11-10 08:17:28.625719045 +0100
@@ -218,7 +218,7 @@
     _count=0;
     _hasInputPngFile=true;
   }
-  const char pngheader[] = {-119,80,78,71,13,10,26,10,0};
+  const char pngheader[] = {0x9,80,78,71,13,10,26,10,0};
   char readbytes[9];
   ifs.read(readbytes, 8);

--- src/formats/yasaraformat.cpp.orig   2016-09-21 21:55:37.000000000 +0200
+++ src/formats/yasaraformat.cpp        2017-11-10 08:07:02.620775489 +0100
@@ -472,7 +472,7 @@

   //  bool hetatom;
   char buffer[32],/*resname[4],*/atomname[5];
-  char double1[8]={0,0,0,0,0,0,-16,0x3f};
+  char double1[8]={0,0,0,0,0,0,0x10,0x3f};
   //   char *str;
   int i,j,/*m,q,*/pos;
   int /*resno,chainNum,link,linktype,*/atoms,element,links/*,chain*/;
```

baoilleach:
Fix #1689. Can you check @newbluemoon?

### DIFF
--- a/src/formats/pngformat.cpp
+++ b/src/formats/pngformat.cpp
@@ -218,7 +218,7 @@ bool PNGFormat::ReadMolecule(OBBase* pOb, OBConversion* pConv)
     _count=0;
     _hasInputPngFile=true;
   }
-  const char pngheader[] = {-119,80,78,71,13,10,26,10,0};
+  const unsigned char pngheader[] = {137,80,78,71,13,10,26,10,0};
   char readbytes[9];
   ifs.read(readbytes, 8);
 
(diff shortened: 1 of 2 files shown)


newbluemoon:
Yes, the patch works! Compilation finishes without any error now. Thank you!

PR #1747, opened 2018-01-08
Fix GCC warnings

baoilleach:
Here are a bunch of commits to fix warnings observed in the core API source files. These warnings are found with -DCMAKE_CXX_FLAGS="-W -Wall". There are still more warnings in other files, but this is a first pass.

Most of these warnings are harmless and relate to signed/unsigned. Here and there I've made a minor mod to the API where a parameter should have been unsigned int in the first place, rather than int, but I don't think there should be anything much controversial in here.

I note in passing that some of the other warnings I've yet to tackle are actual bugs, so it's a useful exercise, especially if we can squash them all and turn on -Werror in Travis.

### DIFF
--- a/include/openbabel/data.h
+++ b/include/openbabel/data.h
@@ -34,7 +34,6 @@ namespace OpenBabel
 
   class OBAtom;
   class OBMol;
-  class OBBitVec;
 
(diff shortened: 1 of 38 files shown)


baoilleach:
Looks like failing at the moment. Will sort out....

PR #1750, opened 2018-01-09
Fix #1684

### LINKED ISSUE #1684: string comparison does not work with numeric sd titles

## Environment Information
Open Babel version: git master
Operating system and version: Ubuntu 17.10

## Expected Behavior
According to [the documentation](http://open-babel.readthedocs.io/en/latest/Command-line_tools/babel.html#string-descriptors), `obabel test/files/filterset.sdf -osmi --filter "title<'129'"` should return 2 molecules, 123 and 126

## Actual Behavior
No molecules are returned

## Steps to Reproduce
    $ obabel test/files/filterset.sdf -osmi --filter "title<'129'"
    0 molecules converted

cowsandmilk:
When figuring out if a filter value is numeric by casting to double, Mac and Linux
set the failbit if the cast cannot be done, so peek returns -1. This checks for
said failbit.

Note, in 1684, the current code appears to work for Noel, so behavior on windows may be different.

### DIFF
--- a/src/descriptor.cpp
+++ b/src/descriptor.cpp
@@ -266,7 +266,7 @@ double OBDescriptor::ParsePredicate(istream& optionText, char& ch1, char& ch2, s
   streampos spos = optionText.tellg();
   optionText >> val;
    //only a number when the param has no additional text or only a closing bracket
-  if(!optionText.eof() && isalpha(optionText.peek()))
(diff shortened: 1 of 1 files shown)


baoilleach:
Thanks. Best fix ever. I looked into this a bit, but couldn't figure it out.

PR #1752, opened 2018-01-11
Gaussformat reading multiple charge modles 2

mmghahremanpour:
SetOrigin() is added.

### DIFF
--- a/include/openbabel/generic.h
+++ b/include/openbabel/generic.h
@@ -1195,9 +1195,9 @@ namespace OpenBabel
       return _PartialCharge.size(); 
     }
     
-    void AddPartialCharge(double q)
+    void AddPartialCharge(std::vector<double> q)
     {
-      _PartialCharge.push_back(q);
+      _PartialCharge = q;
     }
 
     std::vector<double> GetPartialCharge()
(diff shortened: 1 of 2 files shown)


mmghahremanpour:
please let me know if I still need to modify this patch!

ghutchis:
Thanks for the reminder - sorry for the very slow review.

It looks okay. My last concern is that with the current implementation, there's nothing guaranteeing that a molecule will have a full set of partial charges.

I think I'd prefer that the OBPartialCharge class only accept charges as a full vector<double> and checks to make sure it's equal to the number of atoms before storing.

The reader should then push_back() into a temporary vector in the gauss format.cpp code.

This way, if there's somehow a corrupted file (e.g., it got killed by the queuing software before the job finished) the partial charges are still safe - it won't store a set of ESP charges that are only half the molecule.

I realize that's an unlikely case, but I'm trying to think from the library API perspective - you can access a set of vector<double> but set the charges one-by-one. That's asymmetric.

mmghahremanpour:
The code has been updated based on your suggestions.

ghutchis:
Thanks - sorry for the long review process on this one.

mmghahremanpour:
Thank you for merging. 
Would you please review #1753 as well? It is also related to reading Gaussian log files. 

Thank you,

Mohammad

PR #1753, opened 2018-01-12
Fixing minor things for reading ESP from log files

mmghahremanpour:


mmghahremanpour:
Would you please review this patch?

PR #1754, opened 2018-01-15
IsElement implemention, make OBElements::Element enum

dkoes:
This implements an obatom->IsElement(OBElements::Element) method as a more general replacement of the removed IsHydrogen, IsNitrogen etc methods.

I've also made the atomic number constants in OBElements into an enum.  This allows the user to associate a bit of type information with atomic numbers and hopefully write more readable and checkable code.  I did my best to update the bindings as well.


baoilleach:
There are two issues here: the enum and IsElement. I don't either of these changes should be made to the API.

Regarding the function, IsElement(), we should be removing convenience functions, not adding them. This convenience function is trivially implementable in terms of the existing API. The atomic number is get/set via a Get/SetAtomicNum() function. Adding a duplicate function adds nothing to the API, but confuses users. There should only be one way to do it.

Furthermore, this function will likely give rise to inefficient code, where a user/developer chooses IsElement() in favor of GetAtomicNum(). Obviously, to begin with, there are two function lookups instead of one. But rather than accessing the atomic number once, storing it and then switching on it, one could image accessing it multiple times via if IsElement(x) ; else IsElement(y); etc. I have seen examples like this in the OB codebase relating to IsCarbon() and friends.

Regarding the enum, the Get/SetAtomicNum() do not use an enum, nor is the element number stored internally using an enums. Enums and unsigned ints should not be combined as they are intended to be distinct (for example, at some level of warnings, GCC complains). So getting the atomic number via GetAtomicNum() and then passing it to IsElement() currently compiles, but may also be flagged as a warning by the compiler. Sure, the user can cast it, but at that point why isn't it an unsigned int in the first place? More esoterically, I would argue that atomic number is not an enumerated type conceptually; it's an integer. Enumerated types are typically for assigning integers to non-integer items (e.g. days of the week).

include/openbabel/elements.h
```diff
@@ -80,125 +80,245 @@ namespace OpenBabel
...
-    const unsigned int Copernicium = 112; // Cn
-    const unsigned int Nihonium = 113; // Nh
-    const unsigned int Flerovium = 114; // Fl
-    const unsigned int Moscovium = 115; // Mc
-    const unsigned int Livermorium = 116; // Lv
-    const unsigned int Tennessine = 117; // Ts
-    const unsigned int Oganesson = 118; // Og
-
+    enum Element {
+        NonExistent = 0,
```
mwojcikowski: Usually the atomic number=0 is assigned to a Dummy atom

dkoes:
There aren't two function calls and I think you underestimate what an optimizing compiler is capable of, especially considering these are short functions defined in the header file and will be automatically inlined.  I really don't think there is any efficiency issue unless your compiler sucks.

There are issues in using the enum, but it gives the option of type checking element types.  However, I agree that atomic numbers need to continue to be represented as ints so users can create non-physical atoms, so there will be a disconnect somewhere.

I believe this enables more readable, less bug-prone code, but  the effect is slight so I won't advocate for it any more.

baoilleach:
My bad - it is indeed a single call.

However, when it comes to API additions to core classes, I really think we need stronger reasons that those provided.

ghutchis:
I disagree with @baoilleach - I think the improvement in readable code is **_very_** much worth it. I wouldn't mind adding OBAtom::SetElement() as a "symmetry" to OBAtom::IsElement(). I'm inclined to merge that.

Look, the myriad little IsX function calls were tricky because their implementation didn't always match your expectations.

As to the enum, I'm ambivilent.

dkoes:
I'm not overly attached to the enum - it has pluses and minuses.  I'd be happy to go back to the constants, although perhaps we could include the symbols as constants as well?

ghutchis:
I'd personally support symbols as alternates - also seems expressive (and easier to type) (e.g., F over fluorine)

dkoes:
The one issue with the symbols is you run into clashes with residue symbols when building the bindings (which apparently flatten the namespace).  I resolved this by renaming the residue symbols (e.g. resC) as I assume they get much less usage in OpenBabel than elements.

baoilleach:
The choices here set the scene for other API additions/removals. Are you saying that it's okay to extend the API to add more readable functions that duplicate existing functionality (and may lead to less efficient code)?

If you are set on this, please at least consider making the name consistent with the existing Get/Set function; i.e. Has/IsAtomicNum rather than IsElement.

Regarding the elements, again why must we add duplicates? What problem is this solving? We shouldn't just make changes because "I think that would be useful". Either go completely with symbols, or go completely with IUPAC English names. If it turns out that we need both, we can add them down the line. But we can't remove them if we push them out, just in case.

dkoes:
The goal here is to map the API name to how people think.  "Is this atom a carbon" vs "does this class have an atomic number field equal to a predefined constant".  I'm also trying to ease the transition from the helper methods to a more general method.  Along with this, people think of elements using both the names and the symbols, hence the reason to provide both.

Personally, I think usefulness is a valuable trait in an API.

baoilleach:
It might be better to %ignore the element constants with swig and add them
natively in python, etc. It's slow crossing over to C++ for these. This
will also avoid the need for a rename maybe.

baoilleach:
The silence is getting to me. Why not put everything through unchanged, except for the enum? Readability the same, but avoids the need for casting between enums and unsigned ints.

dkoes:
I changed the enum to a bunch of constants.  Mergeable now?

scripts/php/baphpel.php
```diff
@@ -20,7 +20,7 @@
 ********************************************************************/
 namespace baphpel;
 
-include_once 'openbabel.php';
+include_once '/usr/local/lib/openbabel.php';
```
baoilleach: Is this change intentional as part of this commit?

scripts/openbabel-python.i
```diff
@@ -280,6 +280,13 @@ namespace std { class stringbuf {}; }
 %include <openbabel/format.h>
 %include <openbabel/obconversion.h>
 %include <openbabel/obfunctions.h>
+
+//avoid conflicts with OBElement
+%rename(resC) OpenBabel::OBResidueIndex::C;
+%rename(resI) OpenBabel::OBResidueIndex::I;
+%rename(resU) OpenBabel::OBResidueIndex::U;
+
```
baoilleach: This isn't the way I'd fix this, but I'll send in a pull request myself as this has been delayed enough.

include/openbabel/atom.h
```diff
@@ -33,6 +33,7 @@ GNU General Public License for more details.
 #include <openbabel/base.h>
 #include <openbabel/residue.h>
 #include <openbabel/math/vector3.h>
+#include <openbabel/elements.h>
```
baoilleach: I'm guessing this include is unnecessary now.

scripts/openbabel-ruby.i
```diff
@@ -206,6 +207,7 @@ namespace std { class stringbuf {}; }
 %include <openbabel/obconversion.h>
 %include <openbabel/residue.h>
 %include <openbabel/internalcoord.h>
+%include <openbabel/elements.h>
```
baoilleach: Duplicate.

scripts/openbabel-python.i
dkoes: How do you suggest fixing this?  This was the best solution I could come up with given the flattened namespace, but I agree it is unappealing.

scripts/php/baphpel.php
dkoes: No, not sure where that came from

scripts/openbabel-python.i
baoilleach: I would wrap them (i.e. the elements) in #ifndef SWIG, and redefine them 'natively' in each of the bindings. We shouldn't be crossing over to C++ to access the values of constants in the first place. In general, we do this because we're lazy and they would easily go out of synch. Not so for elements->atomic numbers, they are not arbitrary enum values that may change at a future date. For Python, I was thinking of something like a class OBElements with attributes Carbon, etc. OEChem redefines enums/constants as OBElements_Carbon in the various bindings, but I don't like that very much.

ghutchis:
@dkoes @baoilleach - are we resolved on this one? I'd like to merge if we've reached agreement.

dkoes:
Yup, although I was hoping to find the time to do the python native implementation, but that doesn’t have to be part of this.

ghutchis:
@baoilleach and @dkoes - is this possible to merge soon? I like the concept, but I think we want to finalize OB-3 pretty soon.

dkoes:
Fine with me

baoilleach:
See my last comment.

dkoes:
Oh bugger.  The merge pulled in more than the simple fix I was trying to send.  I obviously screwed up somewhere.

ghutchis:
I'm not a Git guru, but this PR may be borked. My usual process when this happens is to close the PR, cherry-pick the needed commits into a new branch and re-submit a new PR.
If there's a more elegant way, go for it.

dkoes:
I think it's fixed (ignore all the commits and just look at the changes).  This is what happens when you do this sort of thing jetlagged at the airport...

dkoes:
(you'll definitely want to merge as a single squashed commit)

scripts/openbabel-ruby.i
ghutchis: I think ruby.i still has the duplicate.

ghutchis:
I'm going to take a look again tomorrow, but i think it resolves the previous concerns @baoilleach - looking at the "changes" link: https://github.com/openbabel/openbabel/pull/1754/files

Issue #1569, opened 2017-05-17
OpenBabel 2.4.1 loses the second molecule in HIN


tulebaev:
;file H2O2_min.hin
mol 1
atom 1 - H  HW - 0  0.7609  -0.3542  -0.7126 1 2 s
atom 2 - O  OW - 0  0.0001  -0.0691  -0.197 2 1 s 3 s
atom 3 - H  HW - 0  -0.7611  -0.3544  -0.712 1 2 s
endmol 1
mol 2
atom 1 - H  HW - 0  0.0001  1.8609  0.0449 1 2 s
atom 2 - O  OW - 0  0.0004  2.8293  0.0259 2 1 s 3 s
atom 3 - H  HW - 0  -0.0005  3.0876  0.9509 1 2 s
endmol 2
[H2O2_min.hin.txt](https://github.com/openbabel/openbabel/files/1008391/H2O2_min.hin.txt)

`babel  H2O2_min.hin  H2O2_min.sdf`

This example with version of OpenBabel-2.3.2 revealed no problems.
I mean OpenBabel-2.3.2 successfully converts both water molecules.
The problem appears in the newer version OpenBabel-2.4.1
This behavior is observed both in Linux and in Windows.

I found that if comment out the lines 137:140 in the file $OpenBabelDir/src/formats/hinformat.cpp

    // clean out remaining blank lines
    while (ifs.peek() != EOF && ifs.good() &&
          (ifs.peek() == '\n' || ifs.peek() == '\r'))
      ifs.getline(buffer, BUFF_SIZE);

the problem is disappeared.

yishutu:
I just created a pull request #1755 to fix this problem. I think it's the conflict behavior between ifs.peek() and ifs.tellg(). I rewrote the tail blank line removing codes to avoid the use of ifs.peek().

PR #1755, opened 2018-01-15
Solving #1569

yishutu:
I just found that the position of ifs.peek() is moved 1 character forward after ifs.tellg() called. And that's  the reason of issue #1569. Therefore I rewrote tail blank line removing codes to avoid use of ifs.peek().

### DIFF
--- a/src/formats/hinformat.cpp
+++ b/src/formats/hinformat.cpp
@@ -135,9 +135,17 @@ namespace OpenBabel
       }
 
     // clean out remaining blank lines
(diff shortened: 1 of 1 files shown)


ghutchis:
Great - do you have a sample hin file for testing?

yishutu:
I just used the HIN file provided from @tulebaev in #1569 as the sample file for testing. It is available here: https://github.com/openbabel/openbabel/files/1008391/H2O2_min.hin.txt

PR #1757, opened 2018-01-16
Fix TESTDIR definition to allow space in path

mcs07:
I'm not sure where this definition is even used? But this change appears to be necessary to compile when there is whitespace in the path to the Open Babel source directory.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -92,8 +92,6 @@ foreach(origtest ${origtests})
   endif()
 endforeach()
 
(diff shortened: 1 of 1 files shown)


baoilleach:
Try removing the line and seeing if the automated build passes.

mcs07:
Seems fine without it.

baoilleach:
Drat - I merged but should have changed the title. Not to worry.

It looks like this was something I added when I rewrote the test suite, but it wasn't needed in the end.

PR #1765, opened 2018-01-24
add Molecule.make2D function

eloyfelix:
I wanted to generate molfile with 2d coordinates from SMILES and I was not able to do it in a nice way using pybel.

The only option I had was doing something like this:

from pybel import _operations
_operations['gen2D'].Do(mol.OBMol)

I think that would be nicer to have a mol.make2D() function.

### DIFF
--- a/scripts/python/examples/testpybel.py
+++ b/scripts/python/examples/testpybel.py
@@ -125,9 +125,10 @@ class TestToolkit(myTestCase):
 
     def testMake2D(self):
         """Test that 2D coordinate generation does something"""
-        mol = self.mols[0]
+        mol = self.mols[1]
         mol.make2D()
-        self.assertNotEqual(mol.atoms[3].coords, (0., 0., 0.))
+        self.assertNotEqual(mol.atoms[2].coords, (0., 0., 0.))
+        self.assertEqual(mol.atoms[2].coords[2], 0.)
 
     def testMake3D(self):
         """Test that 3D coordinate generation does something"""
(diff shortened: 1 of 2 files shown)


baoilleach:
You could have used mol.draw(show=False, update=True).

eloyfelix:
yep, true, but when I tried it before writing this tiny piece of code I had the no coords warning:

import pybel
mol_smiles = pybel.readstring('smiles', "CCCN")
mol_smiles.draw(show=False, update=True)
ob_molfile = mol_smiles.write('mol')


==============================
*** Open Babel Warning  in WriteMolecule
  No 2D or 3D coordinates exist. Stereochemical information will be stored using an Open Babel extension. To generate 2D or 3D coordinates instead use --gen2D or --gen3D.

now I can see that despite the warning the coordinates are set. this is a bit confusing

baoilleach:
That *is* confusing. I'll look into it. Would you still think make2d is needed if this is fixed?

eloyfelix:
In my opinion it looks much clearer .make2D() than .draw(show=False, update=True). 
It should be more efficient(I didn't time it) as there is much more code and checks in the draw function.
So I would add it :)

baoilleach:
Fair enough. Could you remove the comment in the docstring though, as OASA is not used for 2D coordinate generation. (It was once.)

eloyfelix:
done!

baoilleach:
...and done!

Issue #1702, opened 2017-11-24
Possible conflict with scikit-learn in pip version of pybel


hmcezar:
I was having some problems with the sklearn package and even reported an issue in their github:
https://github.com/scikit-learn/scikit-learn/issues/10196

Later we found out that the source of the problem was actually importing pybel, when I installed the `pip` version of openbabel.
If I install openbabel with `conda install -c openbabel openbabel` everything works properly.

I'm not sure what's the problem, but it seems like a problem on the openbabel side (specifically de `pip` version).

baoilleach:
We're aware of the problem and will try to fix in the next release. Could you provide some details of the specific OS, and versions of the various software so that we can reproduce it.

hmcezar:
Sorry for not providing this information initially

- Anaconda information (Python):
```Current conda install:

               platform : linux-64
          conda version : 4.3.30
       conda is private : False
      conda-env version : 4.3.30
    conda-build version : 2.1.13
         python version : 3.5.4.final.0
       requests version : 2.18.4
       root environment : /home/henrique/anaconda3  (writable)
    default environment : /home/henrique/anaconda3
       envs directories : /home/henrique/anaconda3/envs
                          /home/henrique/.conda/envs
          package cache : /home/henrique/anaconda3/pkgs
                          /home/henrique/.conda/pkgs
           channel URLs : https://repo.continuum.io/pkgs/main/linux-64
                          https://repo.continuum.io/pkgs/main/noarch
                          https://repo.continuum.io/pkgs/free/linux-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/linux-64
                          https://repo.continuum.io/pkgs/r/noarch
                          https://repo.continuum.io/pkgs/pro/linux-64
                          https://repo.continuum.io/pkgs/pro/noarch
            config file : None
             netrc file : None
           offline mode : False
             user-agent : conda/4.3.30 requests/2.18.4 CPython/3.5.4 Linux/3.19.0-32-generic debian/jessie/sid glibc/2.19    
                UID:GID : 1000:1000
```

- Operating System: Linux Mint 17.3 Cinnamon Edition 64bits - Kernel 3.19.0-32-generic

- Versions:
The versions that *don't* work together (all using the above Anaconda Python):
```
openbabel                 2.4.1                     <pip>
scikit-learn              0.19.1           py35hbf1f462_0
```
```
openbabel                 2.4.1                     <pip>
[...]

PR #1775, opened 2018-02-10
In the Python bindings, reset the DL open flags after importing _openbabel

### LINKED ISSUE #917: Use of DL flags in Python bindings can cause problems
A bug was reported for scipy at http://projects.scipy.org/numpy/ticket/1148

"""Openbabel Python bindings call
   sys.setdlopenflags(sys.getdlopenflags() | dl.RTLD_GLOBAL)

Obviously, this *will* causes problems down the stream, as symbols of subsequently loaded libraries are no longer isolated...

You can also reproduce the crash without Openbabel,
   python -c 'import sys, dl; sys.setdlopenflags(sys.getdlopenflags() | dl.RTLD_GLOBAL); import foo'

This behavior of Openbabel's is highly suspect IMHO. If it needs to enable this flag, it probably ought to also restore its value after loading whatever needs loading. """

Maybe it's possible to set this flag back to the default after loading various plugins??

Reported by: @baoilleach

baoilleach:
Should have fixed this a long time ago. The DL flags kludge was needed for some platforms or else the plugins didn't load. I can't say whether it's still needed but at least now the code resets the flag value after forcing the plugins to load with OBPlugin::LoadAllPlugins().

I believe this fixes #917, #1222, and #1702. I can't reproduce the original problems though.

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -82,9 +82,11 @@ if (DO_PYTHON_BINDINGS)
               COMMAND ${CMAKE_COMMAND} -E echo "import sys" > ob.py
               COMMAND ${CMAKE_COMMAND} -E echo "if sys.platform.find('linux'\) != -1:" >> ob.py
               COMMAND ${CMAKE_COMMAND} -E echo "  import ctypes" >> ob.py
+              COMMAND ${CMAKE_COMMAND} -E echo "  flags = sys.getdlopenflags()" >> ob.py
               COMMAND ${CMAKE_COMMAND} -E echo "  sys.setdlopenflags(sys.getdlopenflags() | ctypes.RTLD_GLOBAL)" >> ob.py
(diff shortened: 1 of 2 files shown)


baoilleach:
I was pulling my hair out over this failure, which occurred only with the Clang build (on Linux)...but it's right. It's not loading the plugins that requires this to be set, but "import _openbabel" (or it's equivalent) in the openbabel.py. That pulls in the .so file. Still don't understand the underlying reason why, but at least I know where the fix goes now.

ghutchis:
Looks *great*

mwojcikowski:
Yay, no more conflicts with Scipy, etc (basically any fortran code run with f2py). 

I believe it also fixes numpy/numpy#1746

baoilleach:
I keep thinking about this and the difference between the clang and gcc builds. It's nothing to do with the compiler, so it must be something about the linker flags when either the plugins or openbabel.so are being created. I think that is where the real fix should be.

PR #1777, opened 2018-02-13
fix cdxml stereo bonds

JasonYCHuang:
When converting from molfile to cdxml, stereo bonds conversion is in-correct.

This pull request fixes cdxml stereo bonds.

For example, `alanine.mol` is my source file; `alanine_ori.cdxml` is the conversion result from the current master; `alanine_fix.cdxml`  is the conversion result from the pull request.

[alanine.zip](https://github.com/openbabel/openbabel/files/1720262/alanine.zip)

### DIFF
--- a/src/formats/xml/cdxmlformat.cpp
+++ b/src/formats/xml/cdxmlformat.cpp
@@ -371,7 +371,7 @@ bool ChemDrawXMLFormat::WriteMolecule(OBBase* pOb, OBConversion* pConv)
     if (pbond->IsHash())
       xmlTextWriterWriteFormatAttribute(writer(), C_DISPLAY , "WedgeBegin");
     else if (pbond->IsWedge())
-      xmlTextWriterWriteFormatAttribute(writer(), C_DISPLAY , "WedgedHashEnd");
+      xmlTextWriterWriteFormatAttribute(writer(), C_DISPLAY , "WedgedHashBegin");
     xmlTextWriterEndElement(writer());
   }
   _offset += mol.NumAtoms ();


JasonYCHuang:
Would you please review this pull-request?

ghutchis:
I was actually just looking at it. While I agree that the depiction looks better, with something like this, I always check stereochemistry from ChemDraw:
From mol:
screen shot 2018-02-20 at 4 29 06 pm

Current code (clearly weird, but 'S'):
screen shot 2018-02-20 at 4 29 30 pm

Your patch:
screen shot 2018-02-20 at 4 30 06 pm

Now, I agree that the way the wedge is drawn is just plain wrong. And it seems like there's an inversion in the coordinate scheme. But I'm not going to merge this unless you can convince me that your code is producing the correct stereochemistry.

JasonYCHuang:
Thanks for the correction.

I fix the way the wedge is drawn, and it also shows (s), the same as the original one.

image


[alanine_updat.cdxml.zip](https://github.com/openbabel/openbabel/files/1744486/alanine_updat.cdxml.zip)

baoilleach:
Has this been tested over a larger dataset?

JasonYCHuang:
@baoilleach @ghutchis 

I pick 15 molecules with stereochemisty, and compare them.
Please check compare.docx.
[obabel_cdxml_compare.zip](https://github.com/openbabel/openbabel/files/1796427/obabel_cdxml_compare.zip)

In the compare.docx, the left-hand-side column is the original molfiles, and the right-hand-side column is converted cdxmls from this merge request.
I only check cdxml converted from this merge request, and they look good.

JasonYCHuang:
@baoilleach @ghutchis 
Could you review this PR?

JasonYCHuang:
@baoilleach  @ghutchis 

Could you review this PR?

If you need further info, please feel free to let me know.

baoilleach:
Apologies for the delay. Wouldn't a better fix be to leave the bond symbol alone, and just invert the coordinates in the y-axis? After all, it's important for us to get the y axis correct, otherwise, well, everything is going to be upside down.

JasonYCHuang:
@baoilleach Thanks for the feedback.

This PR is only fix the direction of `WedgedHashBond`.
It is an error, and our chemists complain it.

This is nothing to do with coordinate inversion, since the coordinate inversion applies to all stereo and non-stereo molecules.

As to the coordinate inversion, although it makes the graph change, molecules are still valid and the same.
I would like to create another PR to remove this inversion, because it is a different topic from my viewpoint.

What do you think?

baoilleach:
Fair enough, especially if you are going to look at the coordinate inversion in any case.

PR #1790, opened 2018-03-14
Pdbqt fix

dkoes:
master is not generating correct atom types when outputting pdbqt files.  This is because the pdbqt writer works on copies (possibly fragments) of the original molecule.  The copies have their perception flags reset.  By default, nonpolar hydrogens are removed when outputting pdbqt files.  When the aromaticity flags are recomputed on the molecule with the deleted hydrogens, atoms that should be aromatic are not marked as such.

Note that this is a regression from 2.4.  My suspicion is that there is a difference in how the new aromatic typer is dealing with molecules with deleted hydrogens.  I would appreciate it if @baoilleach  could take a look.

I think this fix is good in addition to any fixes that may be made to the aromatic typer as it will prevent an unnecessary recomputation of aromaticity, but would appreciate feedback.  A test case is provided (without the fix the phenyl ring of the tyrosine is not marked aromatic).

### DIFF
--- a/src/formats/pdbqtformat.cpp
+++ b/src/formats/pdbqtformat.cpp
@@ -840,6 +840,10 @@ namespace OpenBabel
     ostream &ofs = *pConv->GetOutStream();
     OBMol & mol = *pmol;
 
+    if(!mol.HasAromaticPerceived()) { //need aromaticity for correct atom typing
+      aromtyper.AssignAromaticFlags(mol);
+    }
+
     if (pConv->IsOption("b",OBConversion::OUTOPTIONS)) {mol.ConnectTheDots(); mol.PerceiveBondOrders();}
     vector <OBMol> all_pieces;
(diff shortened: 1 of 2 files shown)


baoilleach:
Will do.

ghutchis:
This seems like a pragmatic fix to me. I'll leave it to @baoilleach to approve, though.

baoilleach:
Looking at this now...

baoilleach:
I think that the ultimate reason for the regression is that if you delete a bond coming off an aromatic ring (as in this code), you now need to increment the hydrogen count of the two atoms, whereas previously the hydrogen count was "floating" i.e. considered to adjust automatically.

The fix seems fine - I just want to do something first. The second edit you've made should be unnecessary if I "fix" OBMol.Separate(). If the original Mol had its aromaticity perceived, then the Separate components should be returned as having their aromaticity perceived too. Aromaticity perception takes a little while so we should avoid it if possible. So let me do that first, and get back here..

baoilleach:
Actually, it probably makes more sense to merge this, and do the optimization separately.

PR #1793, opened 2018-03-16
Added basic ADF Band and ADF DFTB readers

psavery:
These readers only read atom positions, the unit cell,
and the electronic energy.

They were also added to the generic "out" format so that
they can be automatically detected.

### DIFF
--- a/src/formats/adfformat.cpp
+++ b/src/formats/adfformat.cpp
@@ -1,5 +1,6 @@
 //
 // Copyright (C) 2010 David C. Lonie
+// Copyright (C) 2018 Patrick Avery
 //
 // Molekel - Molecular Visualization Program
 // Copyright (C) 2006, 2007 Swiss National Supercomputing Centre (CSCS)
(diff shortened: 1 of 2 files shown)


psavery:
Hey Geoff. Let me know what you think about this. It is basic, but it serves my own personal needs with it.

In the "out" format, once "ADF" is detected, it now also detects the specific kind of ADF output.

src/formats/adfformat.cpp
```diff
@@ -354,6 +355,284 @@ namespace OpenBabel {
...
+
+          // These are in Bohrs
+          double x = atof(vs[1].c_str()) * BOHR_TO_ANGSTROM;
+          double y = atof(vs[2].c_str()) * BOHR_TO_ANGSTROM;
+          double z = atof(vs[3].c_str()) * BOHR_TO_ANGSTROM;
+          vectors.push_back(vector3(x, y, z));
+        }
+
+        while (vectors.size() < 3)
+          vectors.push_back(vector3(0.0, 0.0, 0.0));
```
ghutchis: Are there always going to be 3 lattice vectors?

src/formats/adfformat.cpp
```diff
@@ -354,6 +355,284 @@ namespace OpenBabel {
...
+              break;
+
+            double x = atof(vs[1].c_str()) * lengthConversion;
+            double y = atof(vs[2].c_str()) * lengthConversion;
+            double z = atof(vs[3].c_str()) * lengthConversion;
+            vectors.push_back(vector3(x, y, z));
+          }
+
+          while (vectors.size() < 3)
+            vectors.push_back(vector3(0.0, 0.0, 0.0));
```
ghutchis: Same here - is there any possibility of a 1D or 2D calculation?

psavery: Same response as above.

ghutchis:
I'd make sure that the vectors are initialized and have some sensible defaults, e.g.,
<100, 0, 0>
<0, 100, 0>
<0, 0, 100>

That way if there are only 1 or 2 vectors, the format won't use uninitialized data.

ghutchis:
At the moment, there's no great way to specify 1D or 2D unit cells. (Sounds like a feature request.)

I'm not sure if client code (e.g., Avogadro) will handle the zero-size vectors, but it's probably a better solution than padding with a giant vector for the 2nd and 3rd lattice vector.

PR #1795, opened 2018-03-17
Use the InChI values for the average atomic mass when reading/writing isotopes

### LINKED ISSUE #1785: Wrong isotope generated from InChI string
- [x ] I believe this to be a bug with Open Babel

## Environment Information
Open Babel version: 2.4.1
Operating system and version: Fedora 27 Linux

## Expected Behavior

% obabel -iinchi Sn120.inchi -omol -OSn120.mol
given file SN120.inchi:
---------
InChI=1S/Sn/i1+1
---------
should produce the file Sn120.mol:
---------

 OpenBabel03041813502D

  1  0  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 Sn  0  0  0  0  0 15  0  0  0  0  0  0
M  ISO  1   1 120
M  END
---------
This is based on the InChI Trust FAQs: https://www.inchi-trust.org/technical-faq/#9.2
"The isotopic shift indicated in the isotopic layer is an integer giving atomic mass relative to the rounded average atomic mass of the element."
For Sn, round(118.710) = 119

## Actual Behavior

However, the command actually generates the following Sn120.mol file:
-------------

 OpenBabel03041813502D

  1  0  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 Sn  0  0  0  0  0 15  0  0  0  0  0  0
M  ISO  1   1 121
M  END
-------------

In comparison, the InChI Trust software does generate the correct output.

## Steps to Reproduce
See above.

baoilleach:
This fixes #1785. InChI uses an offset from the average atomic mass to handle isotopes. Previously we were using Open Babel's values for the average atomic mass, but actually the InChI manual lists values that should instead be used.


baoilleach:
As regards the original bug, the behavior now appears to be correct:


```
>obabel -:"InChI=1S/Sn/i1+1" -omol -iinchi

 OpenBabel03171817122D

  1  0  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 Sn  0  0  0  0  0 15  0  0  0  0  0  0
M  ISO  1   1 120
M  END
```

ghutchis:
Yeah, I mean I stand by using the IUPAC values, but hey, InChI is a standard, so we need to have correct behavior wrt InChI.

hunter-moseley:
Thanks for fixing this!  
The issue is really properly interpreting the InChI isotopomer designation correctly.
Which version of openbabel will have the fix?

baoilleach:
The next version, whatever it may be. But if you compiled it yourself you could manually apply this patch. Or you could use the snap version, though let me know if you want this as I have to trigger the build.

PR #1797, opened 2018-03-18
Fix segfault due to running off the start of an iterator in PDBQT format

baoilleach:
When debugging #1790, I hit a segfault in debug build with MSVC due to running off the start of an iterator.

Note to self: we should really write up the best idiom for deleting a set of bonds.

### DIFF
--- a/src/formats/pdbqtformat.cpp
+++ b/src/formats/pdbqtformat.cpp
@@ -434,34 +434,22 @@ namespace OpenBabel
         best_root_atom=i;
       }
     }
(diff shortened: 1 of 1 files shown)


baoilleach:
Maybe a second pair of eyes would help, @dkoes ?

dkoes:
Wow.  The previous code was impressively convoluted.  The new code is much clearer.

ghutchis:
I'm okay with merging this.

baoilleach:
I was wanted to double check that it was actually doing what I thought it
was. Thanks.

Issue #1768, opened 2018-01-31
Segfault upon reading GAMESS outputs of DFTB3 calculations


sjojungfrun:
I am getting a segfault (core dumped) when reading with openbabel a GAMESS (single core, SMP paralellised) output of a DFTB3 geometry optimisation (file dftb3-2.out attached). For a single point calculation with the same binary, this failure is not found (file dftb3-2-sp.out attached). Besides, when using another GAMESS binary (earlier program version than previous binary, single core IntelMPI pararellised), openbabel does not have problems to read a geometry optimisation with the DFTB3 method (file dftb3-2_oldGamess.out attached). 

This feature is unique of the DFTB3 method. When trying the DFTB one, reading goes smoothly

[attachments.zip](https://github.com/openbabel/openbabel/files/1681190/attachments.zip)

serval2412:
here's the bt:
0x00007ffff7322c44 in OpenBabel::OBAtom::SetPartialCharge (this=0x0, pcharge=0) at /home/julien/projects/openbabel/include/openbabel/atom.h:154
154	      void SetPartialCharge(double pcharge){ _pcharge = pcharge; }
(gdb) bt
#0  0x00007ffff7322c44 in OpenBabel::OBAtom::SetPartialCharge (this=0x0, pcharge=0) at /home/julien/projects/openbabel/include/openbabel/atom.h:154
#1  0x00007fffe4880413 in OpenBabel::GAMESSOutputFormat::ReadMolecule (this=0x7fffe4ab2f30 <OpenBabel::theGAMESSOutputFormat>, pOb=0x55555579f1d0, pConv=0x7fffffffe530)
    at /home/julien/projects/openbabel/src/formats/gamessformat.cpp:478
#2  0x00007ffff75785af in OpenBabel::OBMoleculeFormat::ReadChemObjectImpl (pConv=0x7fffffffe530, pFormat=0x7fffe4ab2f30 <OpenBabel::theGAMESSOutputFormat>)
    at /home/julien/projects/openbabel/src/obmolecformat.cpp:98

The pb is the fact that, on the contrary of the file dftb3-2-sp.out where we got:
TOTAL MULLIKEN
some headings
values...
blank line

in dftb3-2.out, the second occurence with "TOTAL MULLIKEN" contains:
TOTAL MULLIKEN
some headings
values...
DDI: 263640 bytes...

serval2412:
I don't know if the patch will be accepted since it may need an associated test case.
[...]

PR #1798, opened 2018-03-18
Fix#1768: Segfault upon reading GAMESS outputs of DFTB3 calculations

serval2412:
See https://github.com/openbabel/openbabel/issues/1768#issuecomment-374051555
for bt and some explanations

### DIFF
--- a/src/formats/gamessformat.cpp
+++ b/src/formats/gamessformat.cpp
@@ -474,7 +474,10 @@ namespace OpenBabel {
         tokenize(vs, buffer);
         // atom number, atomic symbol, mulliken pop, charge
         while (vs.size() >= 4) {
-          atom = mol.GetAtom(atoi(vs[0].c_str()));
+          int atomNb = atoi(vs[0].c_str());
+          if (!atomNb)
+            break;
+          atom = mol.GetAtom(atomNb);
           atom->SetPartialCharge(atof(vs[3].c_str()));
 
           if (!ifs.getline(buffer, BUFF_SIZE))


baoilleach:
Thanks. Would you consider including a test case? Our test suite is run automatically and will let us know if a future commit would break your code.

serval2412:
I understand but it's another step to do and I don't want to dive in this part.
Don't hesitate to refuse and close the patch if you think it can't be accepted with a test case.

baoilleach:
It can be accepted but it'll just slow down the review, because I'll have to actually look and see what you've done. With a test case, I can just go "fine". :-)

serval2412:
About the test case, you can check the attachment of #1768. Without the patch it indeed segfaults, with it it's ok. Now I recognize that it may not be the right fix (perhaps I'm just hidding the pb).
Also, I checked the existing test cases but I didn't found something enough similar to copy cat it.

ghutchis:
Looks fine for now - it does fix the crash with the #1798 1768.

baoilleach:
Regarding a testcase, something in testbindings.py like (note: untested):
'''
def test1768:
   mol = pybel.readstring("gamout", os.path.join(testfiles, "testfile.gamout.gz")).OBMol
   self.assertEqual(22, mol.NumAtoms())
''''
or if you prefer C++, a similar test in regressions.cpp.

serval2412:
baoilleach: even if I understand the interest of testcases, I'm not very interested in creating some. Fixing a bug is far more motivating.

PR #1800, opened 2018-03-21
When calling OBMol.Separate, preserve whether aromaticity has been perceived

baoilleach:
It's an easy win to preserve whether aromaticity has been perceived when fragmenting a molecule. To do this, I also needed to make sure that this flag was maintained when copying a molecule via "=".


baoilleach:
Note to self: this PR was prompted by @dkoes via #1790.

PR #1792, opened 2018-03-16
Raise a warning when coordinate is NaN

n-yoshikawa:
This pull request is related to #1662.

There is a case when the coordinates calculated by OBBuilder include NaN.
This PR raises a warning when there exists NaN in calculated coordinates.
This does not solve the issue completely, but raising a warning is necessary until it is fixed.

I show the result of the following code. 
```cpp
#include <iostream>
#include <string>
#include <sstream>

#include <openbabel/mol.h>
#include <openbabel/obconversion.h>
#include <openbabel/builder.h>

int main(int argc, char** argv){
  std::string smiles = "FCl(=O)=O";
  std::stringstream ss(smiles);
  OpenBabel::OBConversion conv(&ss, &std::cout);
  conv.SetInAndOutFormats("smi", "sdf");
  OpenBabel::OBMol mol;
  conv.Read(&mol);

  OpenBabel::OBBuilder builder;
  builder.Build(mol);

  conv.Write(&mol);
}
```
Current result:
```

 OpenBabel03161815293D

  4  3  0  0  0  0  0  0  0  0999 V2000
    1.0279    0.0888   -0.0365 F   0  0  0  0  0  0  0  0  0  0  0  0
    2.6179    0.0888   -0.0365 Cl  0  0  0  0  0  0  0  0  0  0  0  0
      -nan      -nan      -nan O   0  0  0  0  0  0  0  0  0  0  0  0
      -nan      -nan      -nan O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  2  0  0  0  0
  2  4  2  0  0  0  0
M  END
$$$$
```

After merging this PR:
```
==============================
*** Open Babel Warning  in Build
  There exists NaN in calculated coordinates.

 OpenBabel03161815253D

  4  3  0  0  0  0  0  0  0  0999 V2000
    0.9442   -0.0573    0.0600 F   0  0  0  0  0  0  0  0  0  0  0  0
    2.5342   -0.0573    0.0600 Cl  0  0  0  0  0  0  0  0  0  0  0  0
      -nan      -nan      -nan O   0  0  0  0  0  0  0  0  0  0  0  0
      -nan      -nan      -nan O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  2  0  0  0  0
  2  4  2  0  0  0  0
M  END
$$$$
```

PR #1801, opened 2018-03-21
Always ensure hybridization

ghutchis:
This addresses the NaN occurring in OBBuilder (e.g., #1792)
because a hybridization might be 0 for a 'strange' atom, e.g.
FCl(=O)=O. As a fallback, assign a hybridization based on the
valence. Not sure it's ideal, but it's better than nothing.

### DIFF
--- a/src/typer.cpp
+++ b/src/typer.cpp
@@ -200,9 +200,25 @@ namespace OpenBabel
     }
 
     // check all atoms to make sure *some* hybridization is assigned
+    int valence;
     for (atom = mol.BeginAtom(k);atom;atom = mol.NextAtom(k))
-      if (atom->GetHyb() == 0)
-        atom->SetHyb(atom->GetValence());
+      if (atom->GetHyb() == 0) {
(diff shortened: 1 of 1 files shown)


baoilleach:
As an aside, I see hybidization as a type of atom typing, so it'd be great to move away from hybridization altogether and just test the structure for whatever feature we are actually looking for. It's not the case so much in the builder, but I think we end up with code where people favor checking the hybridization (with its resultant SMARTS patterns running over the entire molecule) rather than what they really want, like does it have a double bond.

ghutchis:
Agreed, I want to get rid of hybridization - at least the hybridization typer. I'm okay with keeping it as an atom attribute, but I want to minimize the use through the code.

In the meantime (for the builder), this seemed like a good intermediate step - ensure there's always something set. I haven't looked through issues, but I know there have been occasional complaints along this line about NaNs.

Okay to merge?

ghutchis:
Shall we merge this?

Issue #1786, opened 2018-03-05
OBResidue::GetNum() unsigned and negative residue numbers


ntropia2:
When using GetNum() with residues with negative residue numbers, the output is wrong.

- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: (git)
Operating system and version: (all)

## Expected Behavior
Return the correct residue number (negative)

## Actual Behavior
The unsigned version of a negative number is returned (-1 -> 4294967295).
Using signed integers would solve the problem.

## Steps to Reproduce
#!/usr/bin/env python
import openbabel as ob

res = ("ATOM      1  N   MET A  -1     -14.828  18.011  17.067  1.00 23.45           N\n"
"ATOM      2  CA  MET A  -1     -14.103  16.799  16.701  1.00 20.23           C\n"
"ATOM      3  C   MET A  -1     -12.600  17.037  16.703  1.00 19.11           C\n"
"ATOM      4  O   MET A  -1     -11.816  16.089  16.744  1.00 19.81           O\n" 
"ATOM      5  CB  MET A  -1     -14.558  16.295  15.329  1.00 20.29           C\n" 
"ATOM      6  CG  MET A  -1     -15.990  15.795  15.315  1.00 23.10           C\n" 
"ATOM      7  SD  MET A  -1     -16.273  14.605  16.642  1.00 26.64           S\n" 
"ATOM      8  CE  MET A  -1     -15.012  13.399  16.255  1.00 26.91           C\n")

obConversion =  ob.OBConversion()
obConversion.SetInFormat("pdb")
mol = ob.OBMol()
obConversion.ReadString(mol, res)
res = mol.GetResidue(0)
print res.GetNumString()
print res.GetNum()

baoilleach:
What's the range of values that might occur? Any negative number or just -1?

ghutchis:
Yeah, I feel like this has been changed a few times - aren't there cases where this can be a string too?

ntropia2:
According to the PDB specification ([1], page 187), the residue sequence number is defined from column 23 to 26, and can only be an integer, not a string. 
This means that values there should only range from -999 to 9999.

[1] ftp://ftp.wwpdb.org/pub/pdb/doc/format_descriptions/Format_v33_Letter.pdf

baoilleach:
[...]

PR #1804, opened 2018-03-26
Fix 1786 by changing the return value of OBResidue::GetNum()

baoilleach:
API change: return value changed from unsigned int to int to allow handling of potential negative values.


baoilleach:
Fixes #1786.

baoilleach:
Um - not thinking straight. I need to update references to this function before this is ready to merge. There's just a few.

ghutchis:
I think it was originally 'int' and then there was a push for 'type correctness' at one point.

PR #1813, opened 2018-04-09
Add OBMol::SetChainsPerceived(false)

baoilleach:
Mark that chains perception will be required. Toggling this flag is useful if copying a portion of an OBMol with OBMol::CopySubstructure (see #1811).

Personally, I think that all of these perception get/set/has functions should be removed and replaced with an enum, so that it would be UnsetFlag(ChainsPerceived). Or at the least, don't have an unset function, instead have SetChainsPerceived(false). It's silly to have three functions for every bit in an int. In the meanwhile, here's another one...


baoilleach:
Actually - I see that some of the Set functions already take a true/false argument. I think I'll change this one to be the same rather than add another Unset(). I'll update this PR...

ghutchis:
Probably worth doing more API cleanup, but this is fine for now.

PR #1814, opened 2018-04-11
Replace deprecated std::binder1st

serval2412:


serval2412:
Argh, deprecated for me, syntax error for travis because not c++11 ready.
So bad :-(

PR #1818, opened 2018-04-14
Fixes #1282: check format input is provided

### LINKED ISSUE #1282: obabel segfault from invalid command-line arg
The simple command "obabel -i" causes a segfault.

The problem is several cases like this:

    for (arg = 1; arg < argc; ++arg)
      {
       ...
       case 'i':
         iext = argv[arg] + 2;
         if(!*iext)
           iext = argv[++arg];

Note that arg is incremented without checking to see whether it's past argc, so it's accessing past the end of the argv array.  By luck, it's getting a NULL pointer, but iext could point to anything.  There are a number of instances of this in the command-line parsing section.

Reported by: @cjames53

Original Ticket: [openbabel/bugs/902](https://sourceforge.net/p/openbabel/bugs/902)

serval2412:
to avoid a segfault

### DIFF
--- a/test/testbabel.py
+++ b/test/testbabel.py
@@ -105,6 +105,11 @@ class BaseTest(unittest.TestCase):
 class testOBabel(BaseTest):
     """A series of tests relating to the obabel executable"""
 
+    def testNoInput(self):
+        """Ensure that this does not segfault (PR#1818)"""
+        self.canFindExecutable("obabel")
+        output, error = run_exec("obabel -i")
+
     def testSMItoInChI(self):
         self.canFindExecutable("obabel")
         output, error = run_exec("CC(=O)Cl", "obabel -ismi -oinchi")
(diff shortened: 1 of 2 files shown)


serval2412:
Any update here?

ghutchis:
This really should have a test, e.g., in test/testbabel.py.

baoilleach:
I'll look over it this weekend, and see what's going on.

baoilleach:
It all looks good to me. A minor fix to the test code is that as there is no stdin, you just pass a single argument to run_exec. See docs at https://open-babel.readthedocs.io/en/latest/Contributing/Testing.html. Here's what I've got - it segfaults before the patch, but passes after.
```
    def testNoInput(self):
        """Ensure that this does not segfault (PR#1818)"""
        self.canFindExecutable("obabel")
        output, error = run_exec("obabel -i")
```

serval2412:
@baoilleach : Thank you for your feedback, I added the modified test

PR #1820, opened 2018-04-15
Fix#1331: avoid crash with Q-Chem fragment

serval2412:
Test if an atom has been retrieved before calling SetPartialCharge
+ test if the number of atoms expected by Mulliken paragraph corresponds
to the number of atoms retrieved in the previous Standard Nuclear Orientation paragraph


serval2412:
Any thoughts here?

PR #1824, opened 2018-04-18
Add stereo + obfunctions + kekulize to ruby binding

CamAnNguyen:
Add ruby bindings for functions of `kekulize.h` , `obfunctions.h` and related header

### DIFF
--- a/scripts/openbabel-ruby.i
+++ b/scripts/openbabel-ruby.i
@@ -1,4 +1,5 @@
 %module OpenBabel
+%include "std_string.i"
 
 // These fields are renamed to valid constant names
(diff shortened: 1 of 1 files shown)


baoilleach:
Looks fine, but we should really fix OBChemTsfm.Init and indeed any API function that takes a reference to a std::string (or even just takes a std::string).

CamAnNguyen:
I have removed the `OBChemTsfm.Init` part, SWIG works with `std::string` without define the type map

ghutchis:
@baoilleach - I think for now, I'm fine with merging as-is. Agreed?

baoilleach:
Yes.

PR #1826, opened 2018-04-23
Generate useful error messages if plugins can't be found.

dkoes:
Periodically we run into the problem where openbabel can't find the plugin directory.  Instead of generating a useful error message, you usually get something like "cannot open file" and you spend a certain amount of time trying to figure out what is wrong with the file.  Once you realize the problem, you spend a certain amount of time figuring out what the necessary environment variable (BABEL_LIBDIR) is.  

This pull request will output a useful error message describing the problem and the fix when plugins can't be found.  I've also included a test script, but am hoping someone else can figure out how to hook it into cmake (and possibly document the process).


ghutchis:
Well, you're adding a new type of test (a shell-based one) so it'll take slightly more work than just adding another Python script.

baoilleach:
Actually we should already support that. Let me write the docs for you. Also might be good to allow obabel to work from the build dir even without setting Libdir.

baoilleach:
Oh right. I see. You can't change the environment in a Python script. But you can call out to the shell, run a process and check the return value. Let me give it a try.

mwojcikowski:
In Python you could screw up flags as in #1775, though it's global and had to be reverted.

dkoes:
You could set the environment in python, but you would have to make sure this is done before loadallplugins is called and you might poison all future plugin load attempts.  It would have to be a standalone test.

afonari:
This could work too (both py2 and py3 compatible):
```
import subprocess
import os

def test():
    libdir_var = 'BABEL_LIBDIR'
    libdir_val = os.environ.pop(libdir_var, None)

    msg = subprocess.check_output('echo "CCC" | ./bin/obabel -ismi -ocan', shell=True, stderr=subprocess.STDOUT, universal_newlines=True)
    assert 'cannot read input format!' in msg

    if libdir_val:
        os.environ[libdir_var] = libdir_val

test()

```

baoilleach:
Uh, looks like I forgot about this. @dkoes you need to prod us from time to time. Isn't @ghutchis within reach? I'll make this my absolutist priority.

baoilleach:
IMO, this PR is untestable within the current codebase. For example, I can't get the test to pass on a Linux system where Open Babel is already installed. Unless you have another suggestion, I think we should just merge without the test.

dkoes:
On vacation now so can’t look into it, but shouldn’t the current code base fail?

baoilleach:
Minor correction. You're right - I was wrong. Turns out that not having BABEL_LIBDIR set is different from setting it to the empty string. Anyway, I'll send you a PR to update this PR that incorporates @afonari's Python test.

PR #1828, opened 2018-04-26
Check if residue number should be appended to res name during mol2 conversion

### LINKED ISSUE #1819: Conversion to mol2 adds extra "1" to residue name
When converting to mol2 format residue name is appended with "1" each time conversion is performed.

## Environment Information
Open Babel version: latest git
Operating system and version: Linux Mint

## Expected Behavior
Residue name should be written as is

## Actual Behavior
Residue name is appended with "1". On each subsequent conversion new "1" is added.

## Steps to Reproduce
Residue name in original file 1.mol2 is "LIG"

babel -imol2 1.mol2 -omol2 2.mol2    # LIG1  in output
babel -imol2 2.mol2 -omol2 3.mol2    # LIG11  in output
babel -imol2 3.mol2 -omol2 4.mol2    # LIG111  in output

[mol.zip](https://github.com/openbabel/openbabel/files/1911912/mol.zip)

mwojcikowski:
A simple check before appending residue id to residue name. It checks if the exact number is not already at the end of residue name string.

Fixes #1819


baoilleach:
To be honest, I don't like this. It's what I call automagic. It means that the software is making decisions over which the user has no control and which will only occur in particular circumstances. Either we should always add it, or never add it. This may fix the immediate problem, but I predict that at some point someone is going to regard this behaviour as a bug.

mwojcikowski:
Sure, I'm not totally convinced myself. Lets add the option to disable the default behavior instead.

PR #1831, opened 2018-05-02
Fix cdxml upside down format

JasonYCHuang:
This merge request is to fix CDXML inverted y-axis discussed in [PR1777](https://github.com/openbabel/openbabel/pull/1777).
The `Molfile` coordinate system: increase Y, move up.
The `Cdxml` coordinate system: increase Y, move down.
Two systems define opposite Y-axis directions, and result in an inverse during conversion.

Meanwhile, Chemdraw doesn't auto-center the structure, and I add logic to make sure the structure showed in the field-of-view. If the structure position is in a negative region, it hides.

I pick 15 molecules with stereochemisty, and convert them from `mol` to `cdxml` with this PR.
The `compare.docx` in the zip file shows the conversion is good, and no inverse occures.
[openbabel_cdxml_inverse_updated.zip](https://github.com/openbabel/openbabel/files/1967324/openbabel_cdxml_inverse_updated.zip)


ghutchis:
I think the main thing would be to have a test to ensure this doesn't happen again. If you attach some test files (e.g., an MDL file and the expected ChemDraw version - perhaps as saved by ChemDraw itself) I'm willing to write the testing code.

JasonYCHuang:
@ghutchis 

Here are files for testings.

[mol_n_cdxml.zip](https://github.com/openbabel/openbabel/files/1998747/mol_n_cdxml.zip)

I would like to contribute some testings, but have questions.

1.  Is it to re-run `make install` and `make test` after changing something in test files?
It takes a lot of time to re-run them.

2. Could I run one specific test? 
What I know is to run `make test` after `make install`, and this runs all tests.

3. Is there any test example for similar formatting tests which could be a reference to write a test?

baoilleach:
Hi. You can find information on running and adding a test at:
http://open-babel.readthedocs.io/en/latest/Contributing/Testing.html

JasonYCHuang:
@ghutchis @baoilleach 

Test added.
Please check it.

JasonYCHuang:
@ghutchis @baoilleach 

Test added.
Please check it.

PR #1833, opened 2018-05-04
Allow public access to retrieve gradients

ghutchis:
Relocate GetGradientPtr() and GetGradient(atom) for public use, e.g., in Python:
https://gist.github.com/ghutchis/0ecd97aac76b1b3cf99d7113ffb9e376

```
    for atom in mol.atoms:
        # vector objects have to be de-referenced individually (sigh)
        grad = ff.GetGradient(atom.OBAtom)
        print(grad.GetX(), grad.GetY(), grad.GetZ(), sep=',')
```

### DIFF
--- a/include/openbabel/forcefield.h
+++ b/include/openbabel/forcefield.h
@@ -503,21 +503,6 @@ namespace OpenBabel
       }
     }
 
-    /*! Get the pointer to the gradients
-     */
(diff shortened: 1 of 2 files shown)


ghutchis:
I'd love to figure out a better solution for vector3 or double* in Python...

baoilleach:
The Python SWIG interface could add some Python code to create a tuple for you, e.g. called "coords".

PR #1835, opened 2018-05-12
Fix#192: parse and use '-a' flag for obrotate

serval2412:
Only command-line parsing was missing


serval2412:
I forgot to add "-a" in the usage part but before doing this, should we consider the optional flag may be anywhere in the command line or should it be fixed? (I considered it as last argument)

ghutchis:
Thanks, it's a good idea.

It should be considered anywhere in the command line - that's how arguments are handled in any UNIX utility. And yes, I'd like to see the usage part patched too.

serval2412:
@ghutchis: thank you for your feedback.
I changed usage and modified the algo to search the -a and shift values after it to parse values more easily.

PR #1849, opened 2018-06-06
Simplify/fix behavior of OBAtom::GetResidue so that it behaves like other lazy properties

baoilleach:
There are a few related changes here:
1. Move the setting of the HasChainsPerceived() flag into the perception function itself, rather than out here in OBAtom::GetResidue(). What if someone calls the perception function directly? The new behavior is consistent with other perceived properties.
2. Rewrite the function so that it is driven by the HasChainsPerceived() flag, rather than first checking whether the atom has a _residue assigned. This is the behavior I've been fixing for other lazily perceived properties.
3. Remove the code to force (or avoid) perception added by @ghutchis 11 years ago to help Avogadro and others. This looks like a workaround because (2) was not done. How the lazily evaluated system is supposed to work is that you can force or avoid perception using the HasChainsPerceived() flag. Adding or removing an atom will call EndModify() and blow away this flag (I haven't changed this behavior but we could), but the user can store and reset the flag after edits if so desired.
4. Remove the code to wipe the existing _residue if the chains perception returns false. This code doesn't make any sense as it only wipes a single residue. If all of the residues should be wiped then this is the responsibility of the perception code, which is in charge of setting these.

### DIFF
--- a/include/openbabel/atom.h
+++ b/include/openbabel/atom.h
@@ -272,9 +272,6 @@ namespace OpenBabel
       double     GetPartialCharge();
       //! \return the residue which contains this atom, or NULL if none exists
       OBResidue *GetResidue();
-      //! \param perception implies whether chain perception should occur
-      //! \return the residue which contains this atom, or NULL if none exists
-      OBResidue *GetResidue(bool perception);
       //! \return the molecule which contains this atom, or NULL if none exists
       OBMol     *GetParent()        {return((OBMol*)_parent);}
       //! Create a vector for a new bond from this atom, with length given by the supplied parameter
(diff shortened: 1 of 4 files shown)


baoilleach:
Note to self: fails on testbabel.py on the PDBQT test.

baoilleach:
Hi @dkoes. Can you comment on this? This PR causes your PDBQT test case in testbabel.py to fail as follows. The test case has as output the following:
```
REMARK    1  A    between atoms: _1  and  _2
REMARK    2  A    between atoms: _2  and  _3
```
This PR causes this output to change to:
```
REMARK    1  A    between atoms: N_1  and  CA_2
REMARK    2  A    between atoms: CA_2  and  C_3
```
Looks like an improvement to me, but I just want to doublecheck before I edit the test.

dkoes:
Looks better to me as well

ghutchis:
Looks good to me, we'll just need to change the test itself. Thanks Noel!

PR #1853, opened 2018-06-09
Closes #1536: implement periodic boundary conditions

### LINKED ISSUE #1536: Periodic boundary conditions
It would be great if OB could deal with structures and bond perception over periodic boundaries. This has already been implemented by G. Garberoglio in http://software-lisc.fbk.eu/obgmx/ but it would be great to get his patches merged into the standard version.

Reported by: tdaff

Original Ticket: [openbabel/feature-requests/168](https://sourceforge.net/p/openbabel/feature-requests/168)

bbucior:
This PR partially implements periodic boundary conditions in Open Babel by applying the minimum image convention and using `OBUnitCell` data.  Some formats or existing code may not be expecting this behavior, so it is currently hidden behind an `OB_PERIODIC_MOL` flag in `OBMol`.  See also https://github.com/openbabel/enhancement-proposals/pull/6 for more discussion.  Currently this PR is more of a minimum viable product, but over the summer I'm planning to submit additional PR's after combing through more of the codebase.

This PR also adds an option to write bonding information to CIF output, including notation for bonds across periodic boundaries.

Before the code can be merged, there's a handful of details to take care of:

1. There are some merge conflicts (and we should be careful that any changes are compatible with PR#1823).  What's the quickest/cleanest way to address these?
2. Const correctness is tricky: I couldn't keep the `const` keyword on `OBBond::GetLength()` without making a lot of other changes to `OBMol::HasFlag` and some related methods.
3. What is best practice for the argument of `OBGenericData->Clone(NULL)`?  Is it NULL, the OBMol, or something context-dependent?
4. Any other comments or cleanup would also be appreciated.  Thanks!

Special thanks to Giovanni Garberoglio for helpful discussion and his code in <http://software-lisc.fbk.eu/obgmx/>.  I tried to combine the best parts of both of our PBC implementations in this PR.


include/openbabel/mol.h
```diff
@@ -321,6 +326,10 @@ enum HydrogenType { AllHydrogen, PolarHydrogen, NonPolarHydrogen };
     //! \return the size of the smallest ring if a and b are in the same ring, 0 otherwise
     //! \since version 2.4
     int AreInSameRing(OBAtom *a, OBAtom *b);
+    //! \return a pointer to the lattice parameters for periodic systems, or NULL if undefined.
+    //! This is an alternative to
+    //! \verbatim OBUnitCell * pCell = (OBUnitCell * )pmol->GetData(OBGenericDataType::UnitCell);\endverbatim
+    OBUnitCell  *GetPeriodicLattice() const { return _unitCell; }
```
ghutchis: No, please no. As you indicate, there's already a way to do this. I don't want to store a _unitCell for every molecule, when only a small % have them. That's the point of the GenericData system

include/openbabel/mol.h
```diff
@@ -104,7 +104,9 @@ namespace OpenBabel
 #define OB_LSSR_MOL              (1<<20)
   //! SpinMultiplicities on atoms have been set in OBMol::AssignSpinMultiplicity()
 #define OB_ATOMSPIN_MOL          (1<<21)
-  // flags 22-32 unspecified
+  //! Molecule is repeating in a periodic unit cell. See OBMol::_unitCell
+#define OB_PERIODIC_MOL          (1<<22)
```
ghutchis: That's fine - it doesn't require more space...

ghutchis:
Overall, this looks good. Obviously having a periodic UFF would be great.

1. The only way to handle merge conflicts is manually. So take mmcifformat.cpp from master and work in any remaining changes, etc.

2. OBBond::GetLength really needs to be `const`

3. The Clone command is called when copying generic data from one molecule to another. If you're keeping track of atom pointers, etc. you'll need a context-dependent version..

include/openbabel/mol.h
bbucior: Agreed.  A redundant copy of the data can also quickly lead to consistency issues.

bbucior:
Thank you for the feedback.

1. Conflicts resolved
2. One challenge I ran into was that some of the existing code wasn't declared as const.  I can restore const to OBBond::GetLength if some of the other methods are declared as const.  Before making a commit, does this [example change](https://github.com/openbabel/openbabel/files/2104595/fix_const.patch.txt) look reasonable, or are there other concerns about API compatibility, etc?
3. I think I understand.  I don't recall anything seeing anything context-specific in OBUnitCell, but keeping track of the molecule anyway probably wouldn't hurt.  The line I was looking at is `newmol.SetData(parent_uc->Clone(NULL));` (in `OBMol::CopySubstructure`), so I think the relevant OBMol would be `&newmol`.

Making sure the UFF code is PBC-compatible sounds like a great feature.  I will work on that after porting some more PBC features from my research code (such as different formats than CIF) to another PR.

ghutchis:
Okay, this looks okay if you can fix the new set of conflicts. Thanks for your patience - I think this will be really helpful. (I have some ideas about packing molecules using code like this and distance geometry / fragment placement.)

bbucior:
Great!  I resolved the merge conflicts and made sure not to change any of the const declarations in the existing API.  Thanks for your help--I think this is good to go.

bbucior:
@ghutchis I'm going to try and get the periodic UFF code ready as a separate PR next week.  Do you think you'll have time to review this PR before then?

ghutchis:
I'm going to commit this to master - hopefully the merge was successful. (We'll find out once it re-runs tests.)

bbucior:
The build is passing again now that merge conflicts have been resolved and the branch has been updated in light of #1951 and #1958.

ghutchis:
Looks good - I'm going to merge this. Many thanks!

baoilleach:
Just looking at this now in the context of the 3.0.1 release. I see a couple of TODOs in there based on memory allocation and pointers. Do you need/want feedback on these?

bbucior:
Hi @baoilleach hope you are doing well.  Thanks for your interest and offer to help.  Is there any section of code in particular that would be good to clean up?  From what I remember, most of the remaining TODOs were either in SetLength (and related methods) or file formatters that still calculated geometry from raw vector3's instead of atom/bond distances.

I'd be happy to take another look and do some more cleanup, though making Open Babel fully PBC-aware wouldn't be something I could commit to right now.  I'm not currently dogfooding the code for research, and implementing PBC the right way might require some larger architectural discussions to see if there's a way to more generically handle coordinates (perhaps an [OBCoord class](https://github.com/openbabel/enhancement-proposals/pull/6#issuecomment-335043397)?).

baoilleach:
Currently tied up with the release. Let's talk later...

PR #1859, opened 2018-06-28
Coulomb overlap integrals for Gaussian and Slater-type charges

mmghahremanpour:
This patch includes analytical solutions for Coulomb overlap integrals for Gaussian and Slater (1s, 2s, and 3s) orbitals.

### DIFF
--- a/include/openbabel/gaussian_integrals.h
+++ b/include/openbabel/gaussian_integrals.h
@@ -1,0 +1,71 @@
+/**********************************************************************************
+gaussian_integrals.h - coulombintegrals for Gaussian s-type orbitals
+
+Copyright (C) 2018 by Mohammad M Ghahremanpour <mohammad.ghahremanpour@icm.uu.se>
+                      David van der Spoel Group
+                      Uppsala University
+
+This file is part of the Open Babel project.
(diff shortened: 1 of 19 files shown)


mmghahremanpour:
continuous-integration/appveyor/pr fails but it seems that none of the problems are related to coulomb integrals in Coulomb overlap integrals for Gaussian and Slater-type charges-4.

yishutu:
I think the problem is about your coulombtest.cpp. See line 675-679 in the log: https://ci.appveyor.com/project/baoilleach/openbabel/build/1.0.928#L675
Maybe the way of linking is different in the Windows environment. But I am not familiar with it.

dspoel:
Maybe one of the CMakeLists.txt is not entirely correct and the coulombtest should only be compiled for the tests. Alternatively, the integrals are not always compiled in because they are compiled conditionally only.

mmghahremanpour:
I played with CMakeLists.txt and tried different things but with no luck. It is confusing because I can compile it without any linking problem on Mac and Linux machines. Do I need to add the new header (.h) files somewhere in addition to include/openabel/ directory?

PR #1862, opened 2018-07-01
Fixed compiler warnings in libinchi.

dspoel:
We should try to get rid of all compiler warning such that we can require this of all new code as well.
% grep -c warning make.log 
85
It does seem doable.


baoilleach:
Modifications to inchi should be made upstream. Also there is a new version available, probably with new warnings. I note that the new version has a clause saying that modifications are not allowed :-/

baoilleach:
But for sure we should get rid of warnings.

dspoel:
Interesting. Then do we need to include it? Could we have an external library only?

ghutchis:
If you happen to have libinchi installed, you can compile with OPENBABEL_USE_SYSTEM_INCHI 

While I'd be okay with changing compiler flags for the inchi directory (e.g. adding CFLAGS='-w'), I'd really rather have an unmodified inchi. It's not our code.

(I'll note that sending patches to eliminate obvious warnings haven't gone anywhere, sadly.)

dspoel:
Closing this.

PR #1870, opened 2018-07-18
Updates to SMILES options to specify first and last atoms when writing

baoilleach:
Specifying 'f' when writing SMILES allows you to specify the first atom but it also adjusts the H count to facilitate concatenation. I think I'd forgotten this. I've now clarified this in the docstring (also for 'l'), and also added option 'g' that does the same but doesn't adjust the H count.

Note to self: check whether Universal SMILES output is currently using 'f'. If so, it should instead be using 'g'.


baoilleach:
There is a failing clang test on the master now. I think it might have been this. I am unable to reproduce it, for the moment due to a lack of a Mac. Am trying with Linux. We'll see.

PR #1738, opened 2017-12-28
Fix dg stereo

ghutchis:
Work in progress

PR #1875, opened 2018-07-27
Distance geometry method

n-yoshikawa:
This pull request is my update to #1738.


ghutchis:
I can't remember - did we merge this in separately? Certainly many of your fixes were needed in the DG code.

n-yoshikawa:
If I.remember correctly, this PR is just for notifying my progress to others and still have a lot of problems. We should not merge.

n-yoshikawa:
update:
- Use random distance matrix for initialization
- Optimize coordinates in 4D
- Use BFGS implemented in [CppOptimizationLibrary](https://github.com/PatWie/CppNumericalSolvers)
- Introduce new error function instead of swapping atoms

I want to discuss where to put `cppoptlib` directory.

ghutchis:
@n-yoshikawa - is this ready for review?

n-yoshikawa:
Yes, it's ready. I put BFGS library at `include/openbabel/LBFGS.h` for now, but I'm not sure this is the correct directory.

ghutchis:
This looks good - can we move the LBFGS outside of `include/openbabel` - it's not really our header and I don't want to imply that.

If you can change the `#include` lines, I think this is good to merge - even if you implement a better DG method.

n-yoshikawa:
I moved the `LBFGS.h` to `include/`. Is this ok?

Also, I found an issue on treatment of chiral center which may cause segmentation fault, so I want to debug it before merge.

n-yoshikawa:
I solved the chiral center problem, but some of tests failed.

This is possibly because convergence check by `OBDistanceGeometry::CheckStereoConstraints()` does not always agree with the check by canonical SMILES.
Convergence check is another serious problem on coordinate generation, which is discussed before: https://sourceforge.net/p/openbabel/mailman/message/36699127/.

I will try canonical SMILES based convergence check and see how it works.

ghutchis:
I don't think you need a canonical SMILES. As long as the atoms aren't re-ordered, I don't think the SMILES should change order, right?

But I'd like to get a list of some cases where the DG declares success but the canonical SMILES is wrong (i.e., before any MMFF cleanup, etc.) - let's see why the stereo constraints claims to be correct and the SMILES is wrong. That seems like a bug.

n-yoshikawa:
I intended to use canonical SMILES for checking input and predicted chirality as the latest commit. It worked in my local environment, but it didn't in Travis CI. I'm investigating why.

I will prepare some cases where the judgement of DG and canonical SMILES does not match.

ghutchis:
I'd like to merge this, but I'm not sure what to do about the test failures.. Any ideas?

n-yoshikawa:
The test failure is supposed to be caused by the difference between SMILES generate from the predicted molecule (https://github.com/openbabel/openbabel/blob/18b5248b4aca14a5c9f224df08520cc4819dbcbf/src/distgeom.cpp#L918) and SMILES generated from SDF made by obabel. I'm not sure why such a difference exists.

baoilleach:
The difference may be caused by the fact that you are branching off an old version of master. I would rebase on master if I were you, check for the failure, and then force push to this branch. Note that it is a *very* good idea to make a copy of your local git repo before doing this in case you need to revert.

n-yoshikawa:
I tried rebase, but the test still failed.

baoilleach:
I'll take a look.

n-yoshikawa:
I misunderstood the meaning of `SetChiralityPerceived()`. I intended to set chirality based on the current 3D structure instead of cached one, but it was opposite. I replaced it to `StereoFrom3D()`, but the test still fails. Maybe SMILES generation is not dependent on these settings?

In addition to distance geometry related test failures, there are many tests failing. I'm not sure rebase was successful. Closing this pull request and branching from the current master may make problem easier.

https://github.com/openbabel/openbabel/blob/2f51910df2aebbfd1258a2c1100d50a3e1ec7d9e/src/distgeom.cpp#L915

e-kwsm:
> Closing this pull request and branching from the current master may make problem easier.

There is no need to close this PR: `git checkout -B` to reset the branch after backup, and force push. Also, `git bisect` is useful to find the bug—I tried locally but it says 41c83677fd5925275bbfde1bff7299480cd2820b is the first bad commit; perhaps 936db37a51e79a05ddda7fa8350f4c86eb48cabf or c11a198d45ca91d2aee241390614776a7b1a9e28 is.

test/testdistgeom.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Some initial tests based on previous geometry stereo bugs
+        (i.e., OBBuilder can't handle correctly)
+        """
+        self.canFindExecutable("obabel")
+
+        # A series of aromatic strings, which should convert to themselves
+        self.smiles = [
+            'c1ccccc1',  # benzene
+            'C/C=C\C',  # Z-butene
```
e-kwsm: `r'C/C=C\C'` or `'C/C=C\\C'`

test/CMakeLists.txt
```diff
@@ -275,7 +269,7 @@ endif (PYTHON_BINDINGS)
 function(add_depict_test testname obfile image_file ref_image)
   set(depict_dir ${CMAKE_BINARY_DIR}/depict_test)
   add_test(NAME test_${testname} COMMAND ${CMAKE_SOURCE_DIR}/test/test_depiction.sh
-      ${CMAKE_BINARY_DIR}/bin/obabel ${ImageMagick_convert_EXECUTABLE}
+      ${CMAKE_BINARY_DIR}/bin/babel ${ImageMagick_convert_EXECUTABLE}
```
e-kwsm: Should not be modified.

n-yoshikawa: You are right. This file should be fixed.

test/testdistgeom.py
n-yoshikawa: Fixed. Thank you for your comment.

ghutchis:
@n-yoshikawa - you read my mind- last night I was thinking 'let's just have easier tests for now and work on improving'

n-yoshikawa:
I made the test case easier to pass the test for now. I think it is allowed as distance geometry is [described as "unstable" in the document](https://open-babel.readthedocs.io/en/latest/Command-line_tools/babel.html#specify-speed)

I want to improve distance geometry anyway and I want to have advice about how to implement `CheckStereoConstraints()`. I'm currently comparing input SMILES and output SMILES, but it does not seem to be working...

PR #1878, opened 2018-08-06
Update reaction support in ChemDraw

baoilleach:
Adds support for the new handling of reactions, such that mixtures of reactions and molecules will be returned as OBMols.

Also fixes the handling of implicit Hs, as it turns out that there were no test cases for CDX files and it was returning benzene without any hydrogens.

There are a few things that don't work, but they didn't work before either (I believe):
1. Molecules with radicals are not correctly handled (in terms of hydrogens)
2. Only the OBConversion:Convert() interface works. Not ::Read(), and so they cannot be read from Python, etc. I don't know whether this is an underlying problem with reading binary files or just something that Chris never got around to implementing.


ghutchis:
We've had trouble with binary (esp. ChemDraw) in the past. It worked at one point in the distant past, but I don't know when the break occurred.

baoilleach:
In that case, I'll check out old versions and see whether I can pinpoint the breaking commit, because that would save me a lot of trouble.

(And I note the failing test.)

baoilleach:
Note to self: works for 2.3.2 but not for 2.4.0.

baoilleach:
Note to self: the breaking change was https://github.com/openbabel/openbabel/commit/4094c68b87415889150377bd9dc419398ac78d8b#diff-d8fc0ac3a34a3eb74df28348ca3d0100 which ironically was a change to add support for reactions. It's a pretty major change unfortunately, but at least shows that the problem is not in OBConversion (for example).

For the moment, I'll just fix the failing test. But I'll come back to this.

baoilleach:
Ok - I see the problem. The original code (like most formats) had a ReadMolecule that returned a single OBMol. In other words, it could only return a single molecule - never multiple. This is the case for 95% of the formats. For example, the SMILES format only returns a single molecule, and multiple calls iterate through the file. But that's not how CDX files are structured - all the contents must be read at once because they contain references to each other.

It was then changed to return multiple molecules; the mechanism for this is different though, it uses AddChemObject(), which adds it to an OBConversion controlled buffer. Technically, the "fix" I have in mind would have OBConversion::Read() trigger the population of the buffer up-front, and then return values one-by-one. I haven't looked into the specifics - maybe there would need to be a new flag on the format to give Read() a heads-up.

fedorg:
While we are at it, I'm pretty sure the [handling of kCDXProp_Name](https://github.com/openbabel/openbabel/blob/c9c500388dac1469364f778f4f4aa3a6ff7cc7c5/src/formats/chemdrawcdx.cpp#L777) is incorrect, as I have discovered while writing my own little CDX parser built with Kaitai Struct in Python.

The spec [makes a sneaky exception](https://www.cambridgesoft.com/services/documentation/sdk/chemdraw/cdx/properties/Name.htm) for the `Name` prop inside an `ObjectTag` such that the `nStyleRuns` field is completely missing, and the prop contains just raw text data. Thus, the code in question will misinterpret the first 2 bytes of text as `nStyleRuns` when the prop is inside the [`ObjectTag`](https://www.cambridgesoft.com/services/documentation/sdk/chemdraw/cdx/ObjectTag.htm) object.
This has caused my parser to crash and I don't see any special casing in C++ here either.

baoilleach:
It would be great if you could file a separate bug with a testcase.

baoilleach:
For what it does, this is good to merge. I looked into the issue regarding returning multiple molecules one-at-a-time and I couldn't see any way to do it. Not saying there isn't one, but that code is a jungle and I couldn't see my way to it.

PR #1880, opened 2018-08-16
Revert "Updates to SMILES options to specify first and last atoms when writing"

baoilleach:
Reverts openbabel/openbabel#1870 as it appears to cause a test failure with clang


baoilleach:
There's something funny going on here. For the moment, I'll just revert this but this commit isn't the problem.

ghutchis:
Thanks - not sure what the problem is.. it looks fine to me too.

PR #1882, opened 2018-08-29
Update qeq.txt

mkrykunov:
I think that Pb3 (lead) was a typo in the qeq.txt file. So it should be just Pb.


ghutchis:
Yes, indeed. Thanks for the patch!

PR #1888, opened 2018-09-13
ADF DFTB: Add new detection string for ADF 2018

psavery:
For ADF 2018 (which has not yet been released), the string
"|     D F T B     |" is no longer printed in DFTB output. This
is partially because they changed several of their executables to
run via an "ams" executable instead, including the dftb executable.

Hopefully, "DFTB Engine" will suffice instead... it seems to work thus
far.


psavery:
I just added another commit to this to allow for energy reading for DFTB in ADF 2018 as well.

PR #1904, opened 2018-10-25
Added basic ADF 2018 ReaxFF format reader

psavery:


baoilleach:
Would it be possible to add a slightly expanded documentation string, and a testfile/testcase? Also, doesn't a CMakeLists.txt need to updated somewhere?

PR #1909, opened 2018-11-11
Hyperlink DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old links with the `dx` subdomain continue to work, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation by applying this update to all static DOI links.

Cheers!


baoilleach:
Awesome. Much appreciated. Sorting out these papercuts is great.

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #1922, opened 2019-01-15
OBConversion reinitialization for WriteString() and WriteFile()


adalke:
- [ X] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 2.4.90 (HEAD as of 14 January 2019)
Operating system and version: macOS 10.12.2

## Expected Behavior

I expect to be able to write to FPS format using OBConversion without having a segfault. Here are examples of expected behavior:

### Example 1 (expected)
```
>>> import pybel
>>> print(pybel.readstring("smi", "C").write("fps")[:70])
#FPS1
#num_bits=1021
#type=OpenBabel-FP2/1
#software=OpenBabel/2.4.90
```
### Example 2 (expected)
```
>>> import openbabel as ob
>>> conv = ob.OBConversion()
>>> conv.SetInAndOutFormats("smi", "fps")
True
>>> mol = ob.OBMol()
>>> conv.ReadString(mol, "C")
True
>>> conv.WriteString(mol)
'#FPS1\n#num_bits=1021\n#type=OpenBabel-FP2/1\n#software=OpenBabel/2.4.90\n ...`
>>> conv.WriteFile(mol, "example.fps")
True
>>> print(open("example.fps").read()[:80])
#FPS1
#num_bits=1021
#type=OpenBabel-FP2/1
#software=OpenBabel/2.4.90
#source=
#
```

## Actual Behavior

The above code on the current Open Babel HEAD (and with 2.4.1) causes segmentation faults:

### Example 1 (actual)
```
>>> import pybel
>>> pybel.readstring("smi", "C").write("fps")
Segmentation fault
```

### Example 2 (actual)
```
>>> import openbabel as ob
>>> conv = ob.OBConversion()
>>> conv.SetInAndOutFormats("smi", "fps")
True
>>> mol = ob.OBMol()
>>> conv.ReadString(mol, "C")
True
>>> conv.WriteString(mol)
Segmentation fault
```
(Replace the WriteString() with the WriteFile() and there is still a segmentation fault.)

The problem seems to be that WriteString() and WriteFile() on a newly constructed OBConversion() leave the Index at the initial value of 0.

Some of the formats use GetOutputIndex() and, if it is 1, perform an initialization. For the FPS format, the initialization loads the correct fingerprint type. If it is uninitialized then it will deference an unassigned variable, causing the segfault.

[...]

PR #1923, opened 2019-01-16
Improve OBConversion::WriteString() and WriteFile() initialization

adalke:
Otherwise the formats won't initialize correctly. (See issue #1922)

This commit includes a new test program which checks WriteString() and WriteFile() for most of the supported formats.

Without this patch, several of them will fail, and saving to FPS format will segfault.

I wasn't sure if it would be better to put the code into OBConversion::Write(OBBase* pOb, ostream* pos) so that if pos is not NULL, ie, when `SetOutStream(pos, false)` is called, then the index would reset.

I decided against it because I didn't know if people called Write() multiple times with the same output file object, so wouldn't want a reset each time.

There also seems to be a problem with pushInput()/popInput() since those only track the ostream. They should also track and restore the index, yes? That's outside the scope of this patch.

### DIFF
--- a/src/obconversion.cpp
+++ b/src/obconversion.cpp
@@ -994,11 +994,17 @@ namespace OpenBabel {
     {
       StreamState savedOut;
       savedOut.pushOutput(*this);
+      // Set/reset the Index to 1 so that any initialization
+      // code in the formatters will be executed.
+      int oldIndex = Index;
+      Index = 1;
 
       SetOutStream(&newStream, false);
       Write(pOb);
-
       savedOut.popOutput(*this);
+
+      // Restore the Index
+      Index = oldIndex;
     }
 
     temp = newStream.str();
@@ -1034,6 +1040,9 @@ namespace OpenBabel {
       }
 
     SetOutStream(ofs, true);
+    // Set/reset the Index so that any initialization code
+    // in the formatters will be executed.
+    Index = 1;
     return Write(pOb);
   }
 
--- a/src/obconversion.cpp
+++ b/src/obconversion.cpp
@@ -899,14 +899,19 @@ namespace OpenBabel {
 
     if(!pOutFormat || !pOutput) return false;
 
-    SetOneObjectOnly(); //So that IsLast() returns true, which is important for XML formats
-
     // Set the locale for number parsing to avoid locale issues: PR#1785463
     obLocale.SetLocale();
     // Also set the C++ stream locale
     locale originalLocale = pOutput->getloc(); // save the original
     locale cNumericLocale(originalLocale, "C", locale::numeric);
     pOutput->imbue(cNumericLocale);
+
+    // Increment the output counter.
+    // This is done *before* the WriteMolecule because some of
+    // the format plugins initialized when GetOutputIndex() == 1.
+    // This matches the original Convert(), which increments
+    // the count for GetChemObject() before the write.
+    Index++;
 
     // The actual work is done here
     bool success = pOutFormat->WriteMolecule(pOb,this);
@@ -994,16 +999,27 @@ namespace OpenBabel {
     {
       StreamState savedOut;
       savedOut.pushOutput(*this);
-      // Set/reset the Index to 1 so that any initialization
+      // The StreamState doesn't save all of the properties so
+      // do it manually here.
+      
+      // Set/reset the Index to 0 so that any initialization
       // code in the formatters will be executed.
       int oldIndex = Index;
-      Index = 1;
+      Index = 0;
+      
+      // We'll only send one object, so save those properties too.
+      bool oldOneObjectOnly = OneObjectOnly;
+      bool oldm_IsLast = m_IsLast;
+      
+      SetOneObjectOnly(true);
 
       SetOutStream(&newStream, false);
       Write(pOb);
       savedOut.popOutput(*this);
 
-      // Restore the Index
+      // Restore the other stream properties
+      m_IsLast = oldm_IsLast;
+      OneObjectOnly = oldOneObjectOnly;
       Index = oldIndex;
     }
 
@@ -1042,7 +1058,10 @@ namespace OpenBabel {
     SetOutStream(ofs, true);
     // Set/reset the Index so that any initialization code
     // in the formatters will be executed.
-    Index = 1;
+    Index = 0;
+    // We can't touch the Last property because only the caller
+    // knows if the first molecule is also the last molecule.
+    
     return Write(pOb);
   }
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -239,7 +239,7 @@ endif(NOT MINGW AND NOT CYGWIN)
 if (PYTHON_BINDINGS)
   include(UsePythonTest)
   set(pybindtests
-      bindings _pybel example cdjsonformat pcjsonformat)
+      bindings _pybel example cdjsonformat pcjsonformat obconv_writers)
   foreach(pybindtest ${pybindtests})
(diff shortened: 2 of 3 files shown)


baoilleach:
Looks good. WriteString and ReadString unfortunately are historically 2nd class citizens compared to Convert().

My only query is regarding "I didn't know if people called Write() multiple times with the same output file object". This is how one writes out an SDF or SMI file, for example (see the code in pybel.Outputfile.write). Will this still work fine? I'm not sure if a multimol CML file works right now, but that'd be another testcase to consider.

adalke:
Thanks for pointing out that common Open Babel idiom. I had forgotten it.

My patched doesn't touch Write(), only WriteString() and WriteFile(). Trying it out now I see that there is a problem. Here's my test code:
```
>>> import pybel
>>> f = pybel.Outputfile("cml", "spam.cml")
>>> f.write(pybel.readstring("smi", "C"))
>>> f.write(pybel.readstring("smi", "O"))
>>> f.close()
```
On a version of Open Babel without my patch, the output is:
```
<molecule>
 <atomArray>
  <atom id="a1" elementType="C"/>
 </atomArray>
</molecule>
<molecule>
 <atomArray>
  <atom id="a1" elementType="O"/>
 </atomArray>
</molecule>
```
On a version of Open Babel with my patch, the output is:
```
<?xml version="1.0"?>
<molecule xmlns="http://www.xml-cml.org/schema">
 <atomArray>
  <atom id="a1" elementType="C" hydrogenCount="4"/>
 </atomArray>
</molecule>
<?xml version="1.0"?>
<molecule xmlns="http://www.xml-cml.org/schema">
 <atomArray>
  <atom id="a1" elementType="O" hydrogenCount="2"/>
 </atomArray>
</molecule>
```
This is because pybel uses WriteFile() to open the file (which sets Index to 1), then calls Write() to re-use the already open file (keeping the Index at 1).

This tells me two things: the CML format isn't designed to store multiple molecules, and Write() needs to increment the Index.

I don't know how to handle the CML problem. I do know how to fix Write().

I'll update the patch accordingly, and add some test cases with of multi-record output.

baoilleach:
Minor correcction, the CML *is* designed to store multiple molecules, as can be shown at the command-line:
```
C:\Users\noel>obabel -:C -:C -ocml
<?xml version="1.0"?>
<cml xmlns="http://www.xml-cml.org/schema">
 <molecule>
  <atomArray>
   <atom id="a1" elementType="C" hydrogenCount="4"/>
  </atomArray>
 </molecule>
 <molecule>
  <atomArray>
   <atom id="a1" elementType="C" hydrogenCount="4"/>
  </atomArray>
 </molecule>
</cml>
2 molecules converted
```

adalke:
Wow. Okay, digging into the code some more shows that the CML output *never* worked for Write(). The relevant code is the end of the code block:

```
  bool OBConversion::Write(OBBase* pOb, ostream* pos)
  {
    if(pos) SetOutStream(pos, false);

    if(!pOutFormat || !pOutput) return false;

    SetOneObjectOnly(); //So that IsLast() returns true, which is important for XML formats
```

The problem is that SetOneObjectOnly() sets m_IsLast to true, which means it cannot be used to write multiple objects (with a header/footer) correctly.

The CML writer has:
```
        //If more than one molecule to be output, write <cml> at start and </cml> at end.
        if(!_pxmlConv->IsLast())
          {
            xmlTextWriterStartElementNS(writer(), prefix, C_CML, uri);
            uri=NULL;
          }
```
but that test for IsLast() returns m_IsLast which is true, so the !-inversion makes it false, so that branch is not taken.

To make things worse, the footer can only be written if IsLast() is set, which must be done for the last molecule. This means Write() cannot be called until you know if it will be the last molecule to write, and manual users must specify that flag.

That is, I see what you mean by "second-class citizen." The fundamental problem is that format API doesn't have header/footer generation as an independent task from writing a molecule, so there needs to be a way for the write to know if a molecule is the first or last molecule.

It *appears* that the right solution is to:

- remove SetOneObjectOnly() from Write
- document the full Write API so people know when to SetLast()
- update pybel so it delays the last Write() until the close(), which is when it knows if the previous molecule really is the last molecule.

A simple implementation of the last will either break code which expects that it's possible to modify outputfile.write(mol) after writing the molecule (which it can't because the outputfile hasn't yet written it), or require a copy.

A more complex one might default to making a copy, with a flag to enable non-copy behavior when the API caller can promise that the molecule object won't be changed.

I can look at this some more and improve the patch, but need feedback on which route to prioritize.

adalke:
To be clear, my thought is to change this patch so it can be used (even if with effort) to generate a valid CML file using the low-level API (as well as the other formats which don't need a header/footer) and keep the existing pybel code valid for the formats which it does support.

I don't plan to touch the pybel code with this patch. Some other patch will change it to generate valid CML output, which it doesn't currently do.

adalke:
I've updated the patch, and updated the title of this patch to better reflect what it does.

The problem was that OBConversion::Write() always set SetOneObjectOnly() so there was no way it could be used to generate CML output. The CML writer looked at Index and IsLast to see if it was a single-molecule or multi-molecule output. If the former, it wrote <molecule> element. If the latter it wrote a <cml> element containing 2 or more <molecule> elements.

Another problem was that it needs to Index++ before calling WriteMolecule(), as that's the only way for the initialization code check (Index==1) to be correct *and* for GetOutputIndex() to reflect the number of molecules written.

That in turn means that WriteFile() and WriteString() need to set Index=0 before calling Write().

In addition, I updated WriteString() so it saves/restores more state, and I added the new testobconversion_writers.py test program to the list of test cases for "make test".

adalke:
I don't know what to do about the 2 of 3 Travis CI build failures. The output says:
```
181/181 Test #181: pybindtest_obconv_writers ........***Failed  Error regular expression found in output. Regex=[ERROR]  1.54 sec
```
How do I have it report the actual error(s)?

baoilleach:
Travis is simply intended to guard against test failures. To actually deal with them, you need to reproduce the failure locally. There is more info on running the tests at: https://open-babel.readthedocs.io/en/latest/Contributing/Testing.html. This describes how to run an individual test.

fredrikw:
I looked at the test failures briefly and it looks like there are two problems. No1 is with the pointcloud format where the coordinates are different than expected. I guess that this might be due to how the actual vdW surface is generated.

The other problem I get is with testFPS.test_multimol_default where there is a ValueError due to a mix of iteration and f.read() (ValueError: Mixing iteration and read methods would lose data).
The latter I fixed by changing the relevant code to:
```
        with open(filename) as f:
            # Ensure there is a header
            line = f.readline()
            self.assertEqual(line[:5], "#FPS1", line)
            # Skip the rest of the header
            ids = []
            inheader = True
            for line in f:
                if line[:1] != "#":
                    inheader = False
                    self.assertEqual(line.count("\t"), 1, "Wrong number of fields?: %r" % (line,))
                    hex_fp, id = line.rstrip("\n").split("\t", 1)
                    ids.append(id)
                elif not inheader:
                    self.fail("Second header?: %r" % (line,))
            if inheader:
                self.fail("Reached end of file too early, after: %r" % (line,))
```

I think that the logic would be unchanged by this.

With this change and removing the pointcloud-test I could get it to pass all tests locally at least...

ghutchis:
@baoilleach - this seems like a good target to finish merging soon...

baoilleach:
Will do. I had thought it a complex problem, but now that I read @fredrikw's analysis, maybe not so much.

baoilleach:
I'll starting looking into it. Setting up build environment now...it's been a while.

baoilleach:
WIP at https://github.com/baoilleach/openbabel/tree/adalke_writestring.

baoilleach:
Merged as #2176.

PR #1944, opened 2019-03-01
Add support for COF format plus tests

pbecherer:
Add support for COF format (Culgi object file) and add Python tests for converting COF to and from MOL and MOL2 formats using obabel executable


ghutchis:
This looks fine to me. Self-contained changes for a new format. Thanks!

PR #1946, opened 2019-03-04
Refactor python bindings so that openbabel.py and pybel.py are within an openbabel folder

baoilleach:
<strike>[Marked as Work-In-Progress, just to give me a chance to check it all works on Windows...]</strike>

Mainly due to a conflict with PyBEL (https://github.com/openbabel/openbabel/issues/1700, https://github.com/pybel/pybel/issues/217), but also as a clean-up, this PR moves pybel.py and openbabel.py into an openbabel folder. This means that users will have to type:
```
from openbabel import openbabel
from openbabel import pybel
```
In the course of doing this, I also fixed an issue where SWIG wouldn't run due to the format of the eigen_define string, and where the RPATH was messed up on the Python _openbabel.so due to caching.


baoilleach:
Pinging @mcs07. Does this look okay?

mcs07:
Looks good to me at first glance (I haven't tested). 

Although, what do you think about only doing this for pybel? i.e. instead change `openbabel.py` to `openbabel/__init__.py`, so the existing `import openbabel` continues to work as before. That should break less existing code (but there may be other downsides I haven't thought of?)

By the way, if you make the change as you currently have it, wouldn't you also need to change the imports within `pybel.py` itself (currently `import openbabel as ob`)? Either to a relative import (`from . import openbabel`) or absolute (`from openbabel import openbabel`).

mcs07:
Seeing as this is a fairly significant breaking change, it should probably be associated with a equally significant version number increase. It's bound to result in many many messages to the mailing list!

To ease the transition, we could also temporarily add a symlink or duplicate of `pybel.py` in the root `site-packages` directory, and display a deprecation warning if it is imported from there. Ideally it would also gracefully skip installing this file if it already exists from PyBEL.  (I'm not sure how to actually do any of this, I'm just the ideas guy...!)

ghutchis:
Well, it's going to be connected to Open Babel 3, so there's a whole pile of other breaking changes in the API...

ghutchis:
I do like the idea that `import openbabel` is retained and only `import pybel` statements need to change.

baoilleach:
Good point regarding the relative import.

Regarding the other, I'm not keen on dumping everything in __init__.py. I think that's a hack. We should just do it the right way this time, or it'll cause confusion. Someone might know how to import pybel, but be unable to work out how to import openbabel. Or vice versa. Sure, it'll be documented but it should just be the same for both. Yes - it's a change, but I remember the transition from PIL to Pillow was something similar.

As an alternative (if possible - not sure until I try), we could write a warning message telling the user how to change their code if they run "import openbabel".

mwojcikowski:
I generally agree with Geoff, so `import openbabel` and `from openbabel import pybel`. It would be good to have `__version__` parameters of the OB library too.

baoilleach:
(I've just been looking into this a bit, and I can't see any way to raise the warning I suggested, which is a pity.)

As a last argument consider that keeping ``import openbabel`` means that people will try ``import pybel`` and get weird errors if they have pyBEL installed. The choice is between a clear problem for all users (at upgrade time) over a confusing problem for a small number of users (but forever). If the majority prefer the second, then so be it. But I hope everyone who chooses that path will be happy to help users that file issues about this.

mcs07:
I don't have strong feelings about this, I just think it would be good to break as little existing code as possible and `from openbabel import openbabel` also just looks a little bit weird.

But if we do go that route, here's a cool trick to keep `import openbabel` working also, but emit a warning:

#### openbabel/\_\_init\_\_.py

```
import sys
import warnings
from . import openbabel

_thismodule = sys.modules[__name__]

class ObProxy(object):
    def __getattr__(self, name):
        if hasattr(_thismodule, name):
            return getattr(_thismodule, name)
        elif hasattr(openbabel, name):
            warnings.warn('"import openbabel" is deprecated, instead use "from openbabel import openbabel"')
            return getattr(openbabel, name)
        else:
            raise AttributeError

sys.modules[__name__] = ObProxy()
```

It's a bit of a hack, but ["occasionally used and recommended" according to BDFL](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html). (Alas, in Python 3.7 this is much easier, as you can just define `__getattr__` on a module to customise the behaviour and skip the whole proxy class). There may also be a way of doing this by adding a custom finder/loader to `sys.meta_path` as described in [PEP302](https://www.python.org/dev/peps/pep-0302/) and [PEP451](https://www.python.org/dev/peps/pep-0451/) (this is how [`six.moves`](https://pythonhosted.org/six/#module-six.moves) works) but that looks really overly complex to me and in the end might be equally hack-ish...

baoilleach:
Just had a thought last night. How about putting ``from .openbabel import *`` into the ``__init__.py``? This would keep ``import openbabel`` working as now, but also support ``from openbabel import openbabel`` with the idea of moving to that exclusively at some point in the future.

baoilleach:
@mcs07 Our comments crossed. I like your suggestion. Will it work for Python 2 also?

Regarding ``from openbabel import openbabel``, I thought that was a fairly common construction. Alternatively, what about ``from openbabel import ob``? This changes the name of the final import though...

mwojcikowski:
The solution from @mcs07 with warning is cool, and we should actually use it
@baoilleach I think I like the `from openbabel import ob` import the least, the other two are equally fine, so whichever you prefer.

baoilleach:
@mwojcikowski Regarding __version__, I should be able to hack in a call to OBReleaseVersion() following the approach of @mcs07. Sound reasonable?

baoilleach:
Actually, I decided to stick the version string in at configure time, and not to rely on a call to the C++ at run-time. The reason is that if there *is* a version mismatch, it just will cause an import failure and we will never find out the version of the Python bindings. This way, we will know that piece of the puzzle.

From my point of view, I think this is ready for merging.

mcs07:
One last thing to consider is how this impacts the independent installation of the python bindings using the `setup.py` file (e.g. via pip). Do we even want to still support that?

baoilleach:
Well, for Linux it never made a whole lot of sense to me. Does it even work? You'd need to have the same version of OB installed by the distro as the latest version available from PyPI. So I never really got it. Surely it's just a receipe for bug reports and unhappy users.

mcs07:
It should work fine on linux. Version mismatch doesn't seem to be much of an issue given the slow rate of releases. The main use case is for virtualenvs, where multiple python environments can all point to the same OB install. But conda environments now seem to be a more popular way of managing multiple python environments now, and they don't need this as they duplicate the entire OB install into each env.

baoilleach:
It seems to be working now, so it's reasonable to keep it working. I'm not really familiar with it though, so if I get stuck on how to test it I might be back to you...

baoilleach:
It would be great if you could help with this @mcs07. I'm just a bit at sea. Do you think you might have a chance to look at it?

mcs07:
Hi Noel, sorry for not getting back to you about this. I had a quick look this evening and I think I can sort it out. I'll have a go at the weekend and let you know.

mcs07:
OK I had a first go at sorting it out, I created a PR against your PR ;-) See https://github.com/baoilleach/openbabel/pull/1

baoilleach:
Wow - that was quick. Also, I confer upon you the "coding past midnight" badge. :-)

PR #1951, opened 2019-03-17
Tidy setting/unsetting of molecule perception flags

baoilleach:
(Part of the pre-3.0 API cleanup)
The current situation is that some molecular perception flags have an unset method. Others do not. Furthermore, some setters take a boolean for set/unset (full disclosure - this may have been all me).

The latter approach of a single method with a boolean seems reasonable. It reduces the number of methods, reduces the developer error (forgetting to add an unset method), and frankly is easier for users - just one method to do two things. 

This PR gets rid of the Unset methods for molecule flag perception, and ensures that all setters (for perception flags) take a boolean. I've also moved HasFlag and SetFlag from "protected" to "public". I get the idea of encapsulation, but consider the case where the user wants to set a custom flag. I can move it back if someone feels strongly.

If this goes through, I will do the same for OBAtom and OBBond, but I think the majority of perception flags are at the OBMol level.

### DIFF
--- a/include/openbabel/mol.h
+++ b/include/openbabel/mol.h
@@ -101,6 +101,11 @@ namespace OpenBabel
   //! Treat as reaction
 #define OB_REACTION_MOL          (1<<22)
   // flags 22-32 unspecified
+
+#define SET_OR_UNSET_FLAG(X) \
+  if (value) SetFlag(X); \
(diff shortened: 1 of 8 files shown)


baoilleach:
I'm going to assume this is good to go...

PR #1952, opened 2019-03-17
Remove outdated stereo code

baoilleach:
Back in 2009 or before, Tim rewrote the stereo handling with some help from me. The original API was preserved, along with related stereo flags, even though they no longer were being used or worked. This PR removes that code.

It also removes obchiral (which uses the old code - no longer working) and OBMol::IsChiral() which did not do what it says (but could be brought back, with a lot of work, at a later date). I've retained the (incorrect) code of OBMol::IsChiral() in reportformat.cpp and mdlformat.cpp to keep things the same for the moment, but these'll have to be sorted out at some point.

Not strictly part of this PR, but I've gotten rid of the typedef of OBEdgeBase to OBBond.

Removing the unused functions from OBBond has highlighted a few cases of bitrotted code that should be using the new stereo but isn't (and so isn't doing anything at the moment). I've just commented these out for the moment, but these'll have to be sorted out at some point.


baoilleach:
I'm going to assume this is good to go...

PR #1953, opened 2019-03-19
Remove OBBond::GetBO() and SetBO()

baoilleach:
These functions have been deprecated since forever and replaced by SetBondOrder(), GetBondOrder().


ghutchis:
minor conflict, but let's do it.

PR #1954, opened 2019-03-21
Remove OBRandom from the public API

baoilleach:
(Proposed PR as part of the API cleanup)

This removes OBRandom from the public API. It's still there, and included via # include. The changes are a bit hacky for the moment, but removing it gives us some breathing space to fix problems and bring it back (if that's what makes sense). 

So OBRandom will probably come back at some point, but it won't be the same. This code of Roger's is not a general pseudorandom number generator, but an iterator over a fixed (and small) set of intergers in pseudorandom order. The details are actually in a talk that's online from when Roger was in Metaphorics (insert reference in here from future).

This is part of the problem. The other part is that random numbers are being used with a time seed throughout the codebase in a way that hampers reproducibility. When I make a 3D structure a second time, I want to see the same coordinates. We can provide the ability to change the seed, but this should be the default. 

Fixing these problems is a bigger job than there is time for right now. Hiding OBRandom will allow us to make any API changes we need without breaking things.

### DIFF
--- a/include/openbabel/conformersearch.h
+++ b/include/openbabel/conformersearch.h
@@ -445,7 +445,7 @@ namespace OpenBabel {
       std::vector<std::vector <int> > dynamic_niches; //!< The dynamically found niches
       std::vector<int> niche_map;                     //!< Procide the sharing niche index, given the key inddex
       
-      OBRandom unique_generator; //!< A unique random number generator for the whole algo
+      void *d; // Opaque pointer - currently for storing OBRandom* which may be removed in future
       bool use_sharing;		//!< Wether to use sharing or not.
       double alpha_share;	//!< The alpha parameter in sharing function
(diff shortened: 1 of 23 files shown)


ghutchis:
Do you want to add a command-line switch to set the seed? This seems like it's a fairly simple way tweak the randomness (e.g., for unit tests).

ghutchis:
Considering there are better random number generators in C++11 I'm all in favor of this.

baoilleach:
I think it's a case-by-case thing. Being able to specify the seed, or better still using a fixed seed, would solve some of the problem. And in other cases, there's no need for randomness in the first place - a good example is random SMILES - what people really want is to generate a whole bunch of (or all possible) different SMILES, but this can be done without ever using random numbers.

PR #1958, opened 2019-04-06
Remove miscellanous headers from mol.h, atom.h and bond.h

baoilleach:
I saw this at the top of mol.h:
```
// Currently includes many headers for 2.x backwards compatibility
// \deprecated -- this will be cleaned up in 3.0 efforts
//      to improve compile time significantly.
// Only include necessary headers and class declaration stubs.
```
So, that's what I did. I did the same for bond.h and atom.h, and did a little bit of additional triming.

I also moved the globals into wherever their classes are defined (so including oberror.h gives you obErrorLog), and removed the typedefs for OBEdge/NodeBase.

It was a bit more work than I expected, so I should probably have resisted. In any case, from the point of view of a user, they may need to include mol.h, bond.h, atom.h, and obiter.h, where previously mol.h would have sufficed.


ghutchis:
I salute you. This is decidedly needed. I'll see if I have time to run https://include-what-you-use.org to minimize headers too.

baoilleach:
You can't imagine how glad I am you're ok with this :-) This was a real rabbit hole. I'll start working on the merge.

ghutchis:
Oh it definitely needed to happen for 3.0. Gotta dig through my notes and see what else I marked for removal / cleanup.

baoilleach:
Oh no - IWYU could have done most of this for me. Oh well. Anyway, if the automated builds pass, I'll merge what I've already done.

PR #1959, opened 2019-04-07
For Travis builds, include output for build failures

baoilleach:
When a test fails, it would be nice to see the test output, especially for those cases where I'm having trouble reproducing the test failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ before_script:
 
 script:
  - make -j2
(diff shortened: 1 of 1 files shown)


ghutchis:
I like the concept - should we merge regardless of travis failure?

baoilleach:
The failure was on master after having merged the obrandom PR. I then unmerged that PR, and master was happy again.

PR #1970, opened 2019-04-17
Improve python script

e-kwsm:
- `from __future__ import print_function`
- replace deprecated `unittest.TestCase.assertEquals` with
  `assertEqual`, `assertTrue`, or `assertFalse`
- use `obj is None` instead of `obj == None`
- remove modules and functions that are imported but not used


baoilleach:
Thanks.

PR #1971, opened 2019-04-18
Make `pybel.Outputfile` compatible with `with` statment

yishutu:
The `__enter__` and `__exit__` methods were added to the `pybel.Outputfile` class, making it compatible with python's `with` statement. The testing script was also modified.

Example:
```python
with self.toolkit.Outputfile("sdf", "testoutput.txt") as outputfile:
    for mol in self.head:
        outputfile.write(mol)
```


baoilleach:
Great.

PR #1976, opened 2019-04-23
"Bye bye babel" - remove the babel binary

baoilleach:
``babel`` has been officially deprecated since 2.4.0 (2016), and unofficially since the arrival of ``obabel``. This PR removes it.


e-kwsm:
In #1979 I partly replaced `babel` with `obabel` in man files, but there is still `babel.1`, which [is loaded by `obabel.1`](https://github.com/openbabel/openbabel/blob/48387991b746a312dab55cd2ce34c1469ad8351c/doc/obabel.1#L1).
Similarly there is no `doc/obabel.html` but `doc/babel.html`, to which other files (e.g. `obchiral.html`) link.

ghutchis:
Whoops, looks like we need to resolve so that we can merge this..

baoilleach:
I agree that we should update the man pages, the Sphinx docs, etc, but surely this can be merged independently. These are code and build system changes.

baoilleach:
....or at you referring to the conflicting file? If the latter, I'll get on it and sort it out.

ghutchis:
No, I was referring to the merge conflict.

PR #1977, opened 2019-04-25
scripts: Fix Python bindings after move to subdir

Reinis:
There were missed path changes in some places.

Fix tests for BINDINGS_ONLY build (OB_MODULE_PATH override is necessary to point to system installation of OB).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,9 @@ if(NOT MSVC)
     }
    " SCANDIR_NEEDS_CONST)
 
-  set(OB_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/${OB_PLUGIN_INSTALL_DIR}")
+  set(OB_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/${OB_PLUGIN_INSTALL_DIR}"
+      CACHE PATH "Set to system install for bindings only build")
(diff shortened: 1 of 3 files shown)


baoilleach:
This PR goes beyond the description you've provided. You should either provide separate PRs for different changes, or else update the description to describe the multiple changes.

Reinis:
Is something like that ok?

baoilleach:
That's better, but we need to sort out the caching, which I previously reverted as it had broken existing behaviour around a local install. 

If you run "cmake" and only afterwards run "cmake -DCMAKE_INSTALL_PREFIX=whatever", it will appear to build and install fine, but the rpaths will not be set correctly. I spent quite some time tracking down the origin of this regression.

Could you describe what BINDINGS_ONLY is intended to do, as it's undocumented? There must be some way to solve the problem while preserving existing behaviour.

jakeogh:
From: https://github.com/openbabel/openbabel/blob/650b1f829ac47cc84877a330366ec9485da3ec99/CMakeLists.txt#L696 BINDINGS_ONLY is intended to be used to only build the selected bindings and not the rest of openbabel. An example use: https://github.com/gentoo/gentoo/blob/master/sci-chemistry/openbabel-python/openbabel-python-2.4.1-r1.ebuild

ghutchis:
As far as I can tell, this is no longer necessary with 3.2, so I'm closing it. If I'm wrong, please open a new PR. Thanks!

PR #1979, opened 2019-04-26
Update doc [skip ci]

e-kwsm:
- Fix "SEE ALSO"
    - replace `babel` with `obabel`
    - fix self-reference
- Remove leading/trailing white spaces
- Other minor fixes


ghutchis:
thanks, this is very helpful!

PR #1980, opened 2019-04-26
Update URL of specification of MDL MOL [skip ci]

e-kwsm:
http://www.mdl.com is no longer available.
I found https://docs.chemaxon.com and replace with it, but I'm not sure it is appropriate.
I would appreciate feedback.

Also, I updated the url in [`src/formats/exampleformat.cpp`](https://github.com/openbabel/openbabel/blob/48387991b746a312dab55cd2ce34c1469ad8351c/src/formats/exampleformat.cpp), but [`XXXFormat`](https://github.com/openbabel/openbabel/blob/48387991b746a312dab55cd2ce34c1469ad8351c/src/formats/exampleformat.cpp#L40) is not used in anywhere. Is the file required?


mcs07:
Not sure the ChemAxon link is appropriate. It presumably should link directly to Biovia, although I don't think there is a good public link any more. Maybe https://www.3dsbiovia.com/products/collaborative-science/biovia-draw/ctfile-no-fee.html

baoilleach:
I think @mcs07 is right.

e-kwsm:
Thanks, I updated the URL.

Issue #1925, opened 2019-01-24
List of inconsistencies related to valence, charge, aromaticity


ehfassio:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 2.4.1
Operating system and version: Ubuntu 18.04 - 4.15.0-43-generic

I previously reported a possible bug related to the aromatic ring perception algorithm (issue 1920). But, as it was not the only case I found, I decided to perform a broader evaluation of other ligands while using PDB files and to report all the issues at once. The result of this analysis is in the attachment.

As previously, I noticed that OpenBabel 2.3.2 generated a correct output for the molecule I was analyzing, then I also decide to use this version in my tests.

In all tests, I used the openbabel 2.4.1 library (python) for creating an OBMol molecule and then I printed the smiles for each one of the inputs (listed below). To check for possible errors of valence, charge, etc, I used the tool Chemicalize from Chemaxon (the figures in the attachment were obtained with this tool).

First, I extract the ligand from the PDB file (all ligands and ids were informed in the attached file).

Four different inputs were used to create the OBMol object:
1) A .mol file obtained by converting the extracted file using OpenBabel 2.3.2.
2) A .mol file obtained by converting the extracted file using OpenBabel 2.4.1. 
3) The .pdb file. Function used: readfile("pdb", file)
4) The .pdb file as a string. Function used:  readstring("pdb", pdb_block)

Besides it, for each ligand I also evaluated the impact of creating the molecule without hydrogens and adding hydrogens according to some pH value.

In the excel file, you'll find a column with the ligand information, one to indicate if I had used hydrogens and other to inform at which pH the hydrogens were added. Also, four other columns were added (one for each input). The red cells indicate if there is an incorrect aromatic perception.

I believe these results can help you guys to detect what are the possible causes for these errors. I have not uploaded the molecule files because it would be too many files and all of them can be accessed through the PDB site. 

If anything is not clear enough please just let me know and I could help you to reproduce the errors.

Excel file: 
[Analysis_of_the_best_input_Send_to_OpenBabel.xls.zip](https://github.com/openbabel/openbabel/files/2789808/Analysis_of_the_best_input_Send_to_OpenBabel.xls.zip)

baoilleach:
Much appreciated. I'll look into it.

dkoes:
Here's another file with likely a similar issue.  
[...]

PR #1981, opened 2019-05-03
avoid non-finite bond vectors

dkoes:
This is putting lipstick on a pig.  The real issue is hydrogens being added where they have no business (#1925).  When placing a new hydrogen, the GetNewBondVector code assumes a hybridization of 2-6.  Otherwise it returns an infinite vector which gets turned into an atom at the origin resulting in a crazy structure.  This "fix" ensures the returned bond vector isn't horrible if the hybridization state is weird.

An example file is attached.
`babel -isdf 3o8g_uff.sdf -omol2 test2_addH_uff.mol2 -h`

[3o8g_uff.sdf.gz](https://github.com/openbabel/openbabel/files/3143788/3o8g_uff.sdf.gz)


ghutchis:
While I agree there's a deeper underlying problem, this isn't lipstick either - it's basic defensive programming.

I'd love to turn the test file into an actual unit test (e.g., checking if there's an atom v. close to the origin).

@baoilleach and I agree that we'd like all PR to include regression tests and this seems like an important one.

dkoes:
The reason I didn’t bother with a test is because the result still isn’t right (those hydrogens shouldn’t be added).

ghutchis:
I mean, I think we should still add a unit test for "no, there is not a hydrogen at the origin".

Let's take the hydrogens as a separate issue / pull request.

@baoilleach - there's a weird valence issue with this file -- the azide is getting a hydrogen in the middle and at the end. I suspect the problem is that the SDF is mal-formed (i.e., even ChemDraw puts an "h" on the terminal N). However, =N= should never be protonated.

baoilleach:
I will look at it asap. Where's the original from @dkoes?

dkoes:
It's the ligand in the PDB 3O8G: https://www.rcsb.org/pdb/download/downloadLigandFiles.do?ligandIdList=O8G&structIdList=3O8G&instanceType=all&excludeUnobserved=false&includeHydrogens=false

baoilleach:
Open Babel reads this SDF file just like Biovia Draw (I just tested with 2019 version). That is, there is an implicit hydrogen on the central nitrogen. In the other words, the error is in the original file - we read it correctly. 

Again, where did this file come from? The link you provided has an SDF file with "M  CHG  2  21   1  22  -1", which would have given the correct number of hydrogens. The file at the top of the bug report is missing this.

francoep:
That file was created by downloading the file that david linked to, and pushing it through the UFF forcefield in rdkit:
`mol=AllChem.MolFromMolFile(pdbligfile)`
`AllChem.UFFOptimizeMolecule(mol)`

which was then written out to the file attached. The link we provided is the database where the SDF exists. It's a SDF file for a ligand that was deposited in the PDB. Note: that line is present in the original downloaded file, but is then absent after running through rdkit.

baoilleach:
Thanks @francoep. I'm pretty sure that RDKit can roundtrip charges so there's something funny going on there - you might want to report it over at https://github.com/rdkit/rdkit/issues if you can verify that it's an RDKit issue.

dkoes:
This is basic defensive programming - what is needed to get it merged?

baoilleach:
Reading through, @ghutchis asked for a test.

ghutchis:
Yeah, I understand it's not the root cause, but having fixed umpteen of these kinds of bugs, I'd really, *really* appreciate having a new unit test of some sort to make sure this lipstick doesn't come back.

dkoes:
There's a test now... still don't like what hydrogens are added, but they at least don't have crazy bond lengths.

ghutchis:
Much appreciated.

PR #1985, opened 2019-05-13
include babelconfig in vector3.h 

dkoes:
vector3.h includes no other openbabel headers and so OBAPI is not defined.  This is a not a problem with user code - it is reasonable to expect a library's header files to be syntactically correct regardless of what order they are included.


dkoes:
This is really minimal, what is needed to get it merged?

Issue #1990, opened 2019-05-22
Missed ctime header in fpsformat.cpp


orex:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: master
Operating system and version: MinGW windows crosscompiling 

## Actual Behavior

 openbabel/src/formats/fpsformat.cpp: Missed header
`#include <ctime>`

ghutchis:
thanks for the bug report - I'm at a conference right now. if you can make a quick pull request, I'd be happy to merge ASAP.

ghutchis:
Thanks for the fix.

PR #1991, opened 2019-05-23
fpsformat.cpp: compile bugfix header added.

orex:
Issue https://github.com/openbabel/openbabel/issues/1990 fix.


ghutchis:
Thanks so much!

PR #1992, opened 2019-06-04
Address Ubuntu bug in defining python install dir

dkoes:
See https://bugs.launchpad.net/ubuntu/+source/python3-defaults/+bug/1814653
Fix might be slightly fragile - not sure if it would be better to check
for Ubuntu 18.04 directly


baoilleach:
Not your fault but this is ridiculous - they have patched Ubuntu's Python such that installers install to the wrong location?

Anyway, thanks for the workaround.

PR #1993, opened 2019-06-08
Add maeparser support to openbabel 

### LINKED ISSUE #1803: Support for Schrodinger MAE files
- [ ] I believe this to be a bug with Open Babel
- [X] This is a feature request

Schrodinger has helpfully made an open source parser for MAE files available:
https://github.com/schrodinger/maeparser

It would be great to support MAE parsing at baseline using the MAE code.

lorton:
Add fully functional maeparser support for both reading and writing mae/maegz format.  Also added some very basic tests.

maeparser APIs require C++11, so the most controversial change here is enabling C++11 if maeparser is enabled (which it would be by default).  All code seems to build fine with this enabled (at least with modern compilers), so hopefully this will be OK for OpenBabel.

Fixes #1803


ghutchis:
Thanks, this would be great!

C++11 is perfectly fine, but I'm not sure on Boost. We're allowing C++11 for v3 because.. well it's 2019 after all. Why do you need to require Boost? Could you use OBBitVec for now instead?

lorton:
I'm honestly not sure why we went with dynamic_bitset, I'll check with the original developer who made this call probably will have some good reason we didn't use std::vector<bool>.

That said - we also use boost for quite a few various things, like effortless gzip fstreams, parsing doubles efficiently, common string manipulations, their very nice unit test package, etc.  These would be relatively annoying to have to gut out, not impossible, but pretty annoying.  Are you committed at this point to not having boost be a requirement?  I can't imagine you have many users who don't have boost either already installed or trivially available through their package manager.

ghutchis:
A general discussion about boost seems better suited to the developer list.

lorton:
Hi @ghutchis - thanks, sounds good.  Do you think I should start said discussion?

lorton:
@ghutchis - this is now "building" in your CI with my latest commit, but only because it sees no Boost and then skips building maeparser and its tests.  Not sure if that's what you want to do here, but if you want me to make any changes just let me know.

lorton:
@ghutchis - just wanted to ping to see if you've had a chance to think about the ideal way to test this?

ghutchis:
We need to add a Travis build with boost enabled, but I'm merging this for now.

ghutchis:
I'm glad to merge `mae` because I'd love to adopt `coordgen` too. 😄

lorton:
Hey @ghutchis, we could probably help with that - did you end up having a summer of code person work on it, or would it be starting from scratch?

ghutchis:
@lorton - we weren't able to get any interested GSoC students. I opened an issue to discuss this here: https://github.com/openbabel/openbabel/issues/2015

baoilleach:
Aren't the file format extensions, .mae and .maegz rather than .MAE and .MAEGZ? If you can confirm, I'll change the register extension lines to be lowercase.

lorton:
You're correct that it's typically written as .mae and .maegz - but is it
case sensitive in execution?  I was just basing it off the example format
where they use "XXX" to represent your format, so I assumed they wanted
uppercase.

PR #1997, opened 2019-06-24
Add SMILES support for elements specified by 3-digit number, e.g. [#101]

baoilleach:
The original patch is from @rogersayle, which I've restricted a bit. This is is a SMILES extension that is already present in OEChem and RDKit, both undocumented and due to Roger. For at least OEChem, support is present for 1- and 2-digit numbers too.

The primary use case is to support new elements where we don't yet have the symbol, but also to read SMILES written by OEChem/RDKit where *they* don't yet have the symbol.

My own take is that this seems relatively safe, at least if restricted to 3-digits. As an implementation detail, if an atomic number >255 is specified, the SMILES is rejected with a warning (we use a char for the atomic number).

### DIFF
--- a/src/formats/smilesformat.cpp
+++ b/src/formats/smilesformat.cpp
@@ -1677,21 +1677,18 @@ namespace OpenBabel {
         break;
 
       case '#':
-        if (_ptr[1]>='0' && _ptr[1]<='9') {
-          element = _ptr[1]-'0';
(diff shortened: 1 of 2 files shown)


ghutchis:
I think this is a sensible extension. I don't imagine there are many of these SMILES "in the wild" but it's a reasonable syntax.

PR #1998, opened 2019-06-25
PDB and PDBQT Insertion Code Fixes

RMeli:
This PR fixes some problems associated to PDB and PDBQT insertion codes:

* When a `OBMolecule` was copied, PDB insertion codes were lost.
* When a PDBQT was parsed with `parseAtomRecord`, insertion codes were lost.

The `OBMolecule` problem was spotted by @dkoes, which suggested a fix.

### DIFF
--- a/src/mol.cpp
+++ b/src/mol.cpp
@@ -1280,10 +1280,7 @@ namespace OpenBabel
           {
             res = NewResidue();
             src_res = src.GetResidue(k);
-            res->SetName(src_res->GetName());
-            res->SetNum(src_res->GetNumString());
-            res->SetChain(src_res->GetChain());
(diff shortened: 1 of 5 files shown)


src/mol.cpp
```diff
@@ -1280,10 +1280,7 @@ namespace OpenBabel
           {
             res = NewResidue();
             src_res = src.GetResidue(k);
-            res->SetName(src_res->GetName());
-            res->SetNum(src_res->GetNumString());
-            res->SetChain(src_res->GetChain());
-            res->SetChainNum(src_res->GetChainNum());
+            *res = *src_res;
```
dkoes: Note that atoms are _not_ copied here.

PR #2001, opened 2019-06-29
Remove deprecated std::bind1st

e-kwsm:
`std::bind1st` is deprecated in C++11 and removed in C++17.

### DIFF
--- a/src/formats/povrayformat.cpp
+++ b/src/formats/povrayformat.cpp
@@ -523,7 +523,7 @@ namespace OpenBabel
 
         /* ---- Add a pigment - statement for start-atom of bond ---- */
         bond_type = bond->GetBeginAtom() -> GetType();
-        bond_type.erase(remove_if(bond_type.begin(), bond_type.end(), bind1st(equal_to<char>(), '.')), bond_type.end());
+        bond_type.erase(remove(bond_type.begin(), bond_type.end(), '.'), bond_type.end());
         ofs << "\t    pigment{color Color_"
             << bond_type
             << "}" << endl;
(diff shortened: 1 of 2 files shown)


e-kwsm:
`-DCMAKE_CXX_STANDARD=14` is used to pass CI.

e-kwsm:
Minimum required versions (lambda) are
- GCC: [4.5](https://gcc.gnu.org/gcc-4.4/cxx0x_status.html)
- Clang: [3.1](https://clang.llvm.org/cxx_status.html#cxx11)
- MSVC: [2012](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015?view=vs-2019#whats-new-for-c-in-visual-studio-2012)

e-kwsm:
`CMAKE_CXX_STANDARD` is set to 11 to use lambda (see also #2043).

dspoel:
The changes in povrayformat.cpp seem unrelated to the topic of the patch or am I missing something?

e-kwsm:
> The changes in povrayformat.cpp seem unrelated to the topic of the patch or am I missing something?

`std::bind1st` returns `std::binder1st`, both of which are deprecated.

BTW the change was incorrect and now fixed. Let's simply use `std::remove`.

correaa:
Is `bind` necessary at all? It seems that what it is used for the criteria in `std::replace_if` is simply equality. In such case, `std::replace` can be used in place of `std::replace_if`.

e-kwsm:
You are right. There is no necessity to use lambda and `std::replace_if`.

PR #2002, opened 2019-06-29
Add deprecated attribute

e-kwsm:
Define `OB_DEPRECATED` and `OB_DEPRECATED_MSG` to declare classes, functions, and variables are deprecated.

Unless `[[deprecated]]` attribute is effective (C++14 feature) or `SWIG` is defined, these macros are ignored

### DIFF
--- a/include/openbabel/atom.h
+++ b/include/openbabel/atom.h
@@ -335,7 +335,8 @@ namespace OpenBabel
       bool HtoMethyl();
       //! Change the hybridization of this atom and modify the geometry accordingly
       //! \return success or failure
-      bool SetHybAndGeom(int);
+      //! \deprecated This will be removed in future versions of Open Babel
+      OB_DEPRECATED bool SetHybAndGeom(int);
       //@}
 
       //! \name Property information
--- a/include/openbabel/bond.h
+++ b/include/openbabel/bond.h
@@ -94,6 +94,7 @@ namespace OpenBabel
       };
       //! Whether this bond has been visited by a graph algorithm
       /** \deprecated Use OBBitVec objects instead to be fully thread-safe. **/
+      OB_DEPRECATED_MSG("Use OBBitVec objects instead to be fully thread-safe.")
       bool Visit;
 
       //! Constructor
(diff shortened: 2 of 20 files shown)


Artoria2e5:
From the looks of it, the appveyor CI (running some msvc compiler) is extremely unhappy with the way you (un)defined OB_DEPRECATED_MSG; apparently the string failed to disappear for some reason.

Clang is playing fine with the attribute but has troubles with the Python binding.

G++ (5.4.0) expanded the attribute as if it supports it, but actually fails to deal with it. You should probably check for `#if __has_cpp_attribute(deprecated) >= 201309` instead.

e-kwsm:
@Artoria2e5 Thank you for your feedback.

Artoria2e5:
with the "has deprecated" check, the fixed requirement on c++ version can be removed. not that it changes anything...

e-kwsm:
> with the "has deprecated" check, the fixed requirement on c++ version can be removed. not that it changes anything...

The default `-std` option of GCC prior to 6.5.0  is `gnu++98` ([5.5.0](https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/Standards.html#C_002b_002b-Language), [6.5.0](https://gcc.gnu.org/onlinedocs/gcc-6.5.0/gcc/Standards.html#C_002b_002b-Language)).
Travis uses GCC 5.4.0 without `-std`, which defines `__has_cpp_attribute(deprecated)` but requires `-std=c++11` or later to use `[[deprecated]]` attribute (https://wandbox.org/permlink/jcPXyPQ6qqdEO9sd), resulting in CI failure.

OTOH, clang works as expected—`__has_cpp_attribute(deprecated)` is not defined unless C++11 or later is enabled,

e-kwsm:
If GCC is older than 5 or clang than 3.6, short circuit evaluation does not seem to be performed, namely compilation of the following fails:
```cpp
#if defined(__has_cpp_attribute) && __has_cpp_attribute(deprecated) >= 201309L
```

Are such old compilers still supported?

ghutchis:
Thanks for sticking with this - definitely helpful for modern compilers.

PR #2003, opened 2019-06-29
Replace Lw with Lr (lawrencium)

e-kwsm:


ghutchis:
I dunno. On the one hand it's the correct symbol. However, UFF does use the incorrect type name in data files (i.e., for backwards correctness, Lw should be retained).

Is there a particular reason that you want to see the change?

e-kwsm:
Closing since Lw is intended for backward compatibility.

PR #2005, opened 2019-07-08
Make pybel compatible with #1975

yishutu:
I have made some change of `pybel.py` to compatible with function name changes in #1975. Some property names were changed keeping the original name as the alias.
`heavyvalence` -> `heavydegree`
`heterovalence` -> `heterodegree`

### DIFF
--- a/scripts/python/openbabel/pybel.py
+++ b/scripts/python/openbabel/pybel.py
@@ -712,10 +712,10 @@ class Atom(object):
        OBAtom -- an Open Babel OBAtom
 
     Attributes:
-       atomicmass, atomicnum, cidx, coords, coordidx, exactmass,
-       formalcharge, heavyvalence, heterovalence, hyb, idx,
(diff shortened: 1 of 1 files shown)


baoilleach:
Hi. Good fix, but I would suggest some changes.

I don't think we should keep the legacy names. The whole problem is that valence and degree were mixed up. If we keep the legacy names, we perpetuate the confusion.

Also, "def valence" should be renamed to "def degree", and the Atom docstring should be updated.

yishutu:
Thanks for suggestion. So there are the change I have made:

- property name changes
  - `heavyvalence` -> `heavydegree`
  - `heterovalence` -> `heterodegree`
  - `valence` -> `degree`
- old property names will raise `AttributeError` to guide users to use the new property names.
- doc strings are updated

ghutchis:
@baoilleach - ready to go?

PR #2008, opened 2019-07-23
Some tests

dkoes:
I've created a test suite of 14k non-redundant ligands from the pdb.  These are provided as reasonably processed sdf files.  testroundtrip.py verifies that they can be read in and roundtripped (with hydrogens) through sdf, mol2, and pdb file formats without mucking things up.  

Initially there were thousands of failures.  Now there are none.  You can go through all the commits for the details, but here are the bigger changes:

- mol2 file format now outputs formal charges with a UNIT_ATOM_ATTR record (there is an option to disable its output)
- pdb files now output bond orders through the popular but non-standard convention of duplicating CONECT records
- sdf files will only have aliases treated as substructure replacements if the target atom is a wildcard
- numerous adjustments to bond typing rules to make them more conservative

The additional tests add a few minutes to testing, but I think it's very important that we include them.  This sort of large test set is needed.

Plans for the future:
- port python driver to c++ to avoid python dependency
- make things work without having to roundtrip the molecules with hydrogens
- compare protonation states to rdkit, find and fix errors
- additional file formats for roundtripping


ghutchis:
Thank you! Much appreciated to have more roundtrip tests.

IIRC, there was a long debate about the whole CONECT thing and I believe the decision was that Open Babel shouldn't output bond order via CONECT by default (i.e., non-standard files). I *thought* that the PDB writer was supposed to have an option to do that, but I don't see it in the code (e.g., -xb = write bond orders).

dkoes:
I'm going to come down pretty strongly on the default including bond orders.  Messing up molecules is never desirable.  I put in an option - if needed it can be reversed and set during testing.

ghutchis:
Please post to the list. I happen to agree with you based on the prevalence of the practice, but consensus was pretty strong against the practice.

I'll search carefully for the previous discussion, but the big concern was that these are not standard PDB files and we should never be writing non-standard files. 

Read what you can, but always write standard files.

ghutchis:
@baoilleach - do you remember the debate about PDB bond orders via CONECT records?

dkoes:
The are not nonstandard files.  There is nothing in the standard that says bonds can't be duplicated: http://www.wwpdb.org/documentation/file-format-content/format33/sect10.html#CONECT

ghutchis:
I appreciate your point, but:
1) I'm not going to merge this without on-list discussion. There has been much debate on this off-and-on for ages, and I've ended up dealing with bug reports and patch requests going both ways. Unless there's public non-GItHub discussion about an enhancement change, someone is bound to get upset.
2) While I can't immediately find the thread, I remember the point being made that the PDB itself was purifying files to remove any use of multiple CONECT records. Moreover, I believe there were cases of software behaving badly to such files (from Open Babel).

My point is - while I like the 'extension', I've had to deal with this debate for ages. I'm not making a change without people having the opportunity to weigh in. If there's consensus (or apathy) to the change, no problem.

baoilleach:
"You can go through all the commits for the details." Fair enough, but there must be 20/30 commits there and the PR description above is minimal. If I am to review it, it will have to wait until the start of September, when I have some time off. Alternatively, either forget about me reviewing it, or else break down the PR into manageable chunks that are orthogonal.

dkoes:
As long as we can get the bug fixes in 3.0, I'm fine waiting (although I don't think it is as bad as it looks).

ghutchis:
Looks like on-list discussion is generally in favor of default to multiple-bond CONECT records if there's an option to write 'traditional' ones too.

I'm not capable of resolving these conflicts @dkoes - but can review this week if you can handle the insertion code changes in `pdbqtformat.cpp`.

ghutchis:
I don't know what all this 2014-2015 stuff is. I'm now having trouble reviewing the specific commits of this PR.

baoilleach:
Just a heads up that after tomorrow I won't have time to review this.

scripts/openbabel-python.i
```diff
@@ -28,6 +28,7 @@
 #include <openbabel/reactionfacade.h>
 #include <openbabel/residue.h>
 #include <openbabel/internalcoord.h>
+#include <openbabel/bondtyper.h>
```
ghutchis: Why would you need the bond typer class in Python?

ghutchis:
There are a *lot* of changes.

Please, I don't want to deal with another 'mega-patch' again - many of these could have been broken into separate pieces (e.g., thread local pieces) for easier review and merging.

dkoes:
Sure, but I started by creating the test set.  Any of the incremental changes were not sufficient to  fully resolve all the failures.

Nonetheless, I will try to do better in the future.  I have learned that I need to rebase frequently to avoid headaches.

e-kwsm:
I wonder why such many files are added to this repo. I don't know where they come from, but how about something like
```sh
wget somewhere/archive
tar -xf archive
# test
```

dkoes:
My preference is to keep the tests independent files rather than an archive.  Otherwise it will be a pain for people to modify/extend the test sets - every small change will require updating the tarball.

ghutchis:
I'm going to merge this for 3.0. I think we probably need to revisit having a separate (larger) test set for cases like this, MMFF94 validation, etc.

I'm going to remove bondtyper from the script bindings unless someone can explain why it's crucial.

dkoes:
According to the commit that added the bondtyper.i include, it was needed to compile successfully with RUN_SWIG, but I just checked and compiles just fine without it now.  :man_shrugging:

baoilleach:
Why are there RDKit related files in there?

dkoes:
> Why are there RDKit related files in there?

I intended to do a comparison in protonation predictions between the toolkits so we could at least have tests for the unambiguous cases, but haven't been able to find the time to do so.

PR #2010, opened 2019-07-30
H vector fix

dkoes:
This is an alternative fix to generating crazy hydrogens.  Instead of defaulting to a hybridization 2 kind of bond, if the builder doesn't think there is a reasonable bond vector, then the hydrogen isn't added.  I've also updated the documentation for GetNewBondVector to state that a non-finite vector can be returned if there isn't a reasonable location for another atom.

This restores the old behavior of GetNewBondVector (return non-finite bond vectors), which is good in the off chance anyone was relying on it while still fixing the problem.  In the corner case that a molecule has multiple conformers, some where the hydrogen can be added and some where it can't the hydrogen is added at (0,0,0), but I'm not sure this can actually happen.

This also comes with updated hydrogen testing code.  The goal is to eventually always perform consistent and reasonable protonation and this PR is a baby step in that direction.

Incidentally, the problem with the test molecule is that it has had the formal charge information stripped from the azide.  In my perfect world, AddHydrogens would also update formal charges if doing so resulted in a more reasonable molecule than adding a hydrogen.

### DIFF
--- a/include/openbabel/builder.h
+++ b/include/openbabel/builder.h
@@ -74,7 +74,8 @@ namespace OpenBabel
       void LoadFragments();
       std::vector<vector3> GetFragmentCoord(std::string smiles);
 
-      /*! Get the position for a new neighbour on atom.
+      /*! Get the position for a new neighbour on atom.  Returns
+       * non-finite vector if there is no reasonable location.
        *  \param atom Atom for which we want a new neighbour location.
        *  \returns The position for the new atom.
        */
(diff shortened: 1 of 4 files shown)


ghutchis:
The commit looks fine - looks like there are some Windows compiler errors though:
https://ci.appveyor.com/project/baoilleach/openbabel/builds/26342761/messages
> 'vcnts' : unknown size test\addhtest.cpp 44   test_runner
> expected constant expression test\addhtest.cpp 44  
> cannot allocate an array of constant size 0 test\addhtest.cpp 44

Artoria2e5:
The variable-length array bites again!

ghutchis:
Thanks - just getting back to check on these things.

PR #2012, opened 2019-08-03
Do not stoke around the text

Artoria2e5:
The current version of OpenBabel adds a stroke around the text, creating an awkward "rounded bold" effect that looks worse than both regular text and bold text. This commit removes such strokes.


ghutchis:
Do you think you can post a before/after graphic?

ghutchis:
The patch looks fine, but it would help to see the difference, thanks!

Artoria2e5:
before
image


after
image

If the text looks too thin i can throw a `style="font-weight:bold"` in there.

ghutchis:
Yes, I think the bold is warranted.

PR #2019, opened 2019-08-28
WIP: Ring loop builder

ghutchis:
Currently work in progress - form ring loops for generic rings that aren't matched in the fragment database.

Previous code will build a straight chain and then a very, very long bond for the ring closure. This code still has serious bugs, but at least makes loops.

Screen Shot 2019-08-28 at 11 47 26 AM


ghutchis:
I'm closing this in favor of another branch.

PR #2022, opened 2019-08-31
Add force field support for dielectric constants in charge terms.

ghutchis:
Needs unit tests, but supported for MMFF94, GAFF, and Ghemical
force fields. Enables minimization and energy evaluation with
decreased electrostatic components (e.g., set epsilon to 500)
or dielectric of different solvents.


ghutchis:
Unless @baoilleach has an objection, I'm going to merge this - minor change, but makes some FF calculations better.

PR #2029, opened 2019-09-09
Substitute deprecated std::bind1st with std::bind`

RMeli:
This PR gets rid of the deprecation warnings concerning `std::bind1st` and `std::binder1st`. They are removed in favour of `std::bind` and `std::placeholders`.


RMeli:
Wooops, I searched among issues but not on PRs and completely missed PR #2001... You can close this PR if the other one is preferred.

RMeli:
I think the CI failures are caused by the fact that C++11 is needed.

PR #2031, opened 2019-09-11
Add a warning message when both -p and -h options are set

yishutu:
When both of `-p` and `-h` options are set, the `obabel` will behave as adding all implicit atoms without considering the pH environment, just like using `-h` only.
This PR adds a warning message to warn users when both of these options are set.


ghutchis:
Great idea, thanks!

PR #2035, opened 2019-09-20
Fixed alias expansion for files with multiple aliases

### LINKED ISSUE #2021: Weird regression in chemdraw handling between 2.3.2 and 2.4.90

## Environment Information
Open Babel version:   2.3.2 and 2.4.90
Operating system and version:  Centos 7.4

## Expected Behavior
under openbabel 2.3.2, using the file attached below, when I run obabel, I get the following output:
`% obabel -icdx 60.cdx -ocan -xr`
`Reading 0x1851180`
`Reading 0x1851180`
`Done reading 0x1851180`
`Reading 0x1851180`
`Done reading 0x1851180`
`Done reading 0x1851180`
`N#Cc1ccc(cc1)C(c1ccccc1)OC1CC2CCC(C1)N2C	AHN 2-020.cdx`
`1 molecule converted
`
This SMILES string is the same as the molecule drawn in the Chemdraw file, and the name provided is the actual internal name of the molecule in the file.

## Actual Behavior
If I run the same command using openbabel 2.4.90 (the version I pulled from git as current this week, on the other hand, I get:
`% obabel -icdx 60.cdx -ocan -xr`
`*c1ccc(cc1)C(c1ccccc1)OC1CC2CCC(C1)N2C`
`1 molecule converted
`
Which is not the same molecule as is depicted in the Chemdraw file; the cyano group has been silently dropeed.  (I have 537 of these old chemdraw files.  When I attempt to convert them to SMILES many of them lose various functional groups, sometimes with a warning that an alias has failed to be converted, but many times as here there is no warning aside from the asterisk in the output).  Also, the molecule's name is no longer being extracted from the file as it was in 2.3.2.  This seems to be a regression in behavior between 2.3.2 and 2.4.90.  Thank you for your attention.

## Steps to Reproduce
[60.cdx.gz](https://github.com/openbabel/openbabel/files/3561276/60.cdx.gz)

fredrikw:
This should fix #2021. Problem was that alias expansion was performed while looping over the atoms.


ghutchis:
Looks good to me.

PR #2038, opened 2019-09-20
Fix reading of uninitialized data.

dkoes:
Valgrind was reporting Conditional jump or move depends on uninitialised value(s)
I am pretty sure that initializing to zeroes is the right thing to do,
but don't really grok the kekulize code so perhaps this is a symptom of
a larger problem.  It's certainly an improvement over reading
uninitialized values.

### DIFF
--- a/src/kekulize.cpp
+++ b/src/kekulize.cpp
@@ -24,6 +24,7 @@ GNU General Public License for more details.
 #include <openbabel/obiter.h>
 #include <openbabel/kekulize.h>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


baoilleach:
Good find. Do you still have the test case? I'd like to see where in the code it was accessing the value, as it may be a logic problem.

dkoes:
[rec.pdb.gz](https://github.com/openbabel/openbabel/files/3637939/rec.pdb.gz)

valgrind babel rec.pdb.gz rec.mol2

PR #2036, opened 2019-09-20
Bumping the major version requires more changes throughout the library.

baoilleach:
Here are my notes from the maintenance repo:
```
On going from 2.x to 3.x, for example, you should make a few more changes:

1. CMakeLists.txt::
   
   set(OB_INCLUDE_DIRS "include/openbabel-3.0") in CMakeLists.txt
     and
   set(OB_EXPORTS_FILE "${openbabel_BINARY_DIR}/OpenBabel3_EXPORTS.cmake")   

2. The package config section::

          configure_file(${openbabel_SOURCE_DIR}/openbabel-3.0.pc.cmake
            ${openbabel_BINARY_DIR}/openbabel-3.0.pc @ONLY)
          install(FILES ${openbabel_BINARY_DIR}/openbabel-3.0.pc
            DESTINATION ${LIB_INSTALL_DIR}/pkgconfig)

  and::

          git mv openbabel-2.0.pc.cmake openbabel-3.0.pc.cmake

3. dlhandler_win32.cpp::

    handle = GetModuleHandle("openbabel-3.dll");//CMake/VC++2008 build will use this.

4. src/CMakeLists.txt::

  # Set the output name to openbabel-3 (preserves old behaviour). Is this desired?
  set_target_properties(openbabel PROPERTIES
    OUTPUT_NAME openbabel-3)
  remove_definitions(_UNICODE UNICODE)
   
5. Then there's a whole bunch of lines in CMakeLists.txt starting with::

     # Create the Config and ConfigVersion files in the build directory, useful to
  
   and the associated files::

        git mv OpenBabel2Config.cmake.in OpenBabel3Config.cmake.in
        git mv OpenBabel2ConfigVersion.cmake.in OpenBabel3ConfigVersion.cmake.in
```

PR #2040, opened 2019-09-20
Remove minor version from some names

baoilleach:
A follow-up to #2036. From my brief Googling, I think the .pc file is now named as it should be. And I left the hyphen out of the include folder - I don't think it's usually there. My notes now read:

```
On going from 2.x to 3.x, for example, you should make a few more changes:

1. CMakeLists.txt::
   
   set(OB_INCLUDE_DIRS "include/openbabel3") in CMakeLists.txt
     and
   set(OB_EXPORTS_FILE "${openbabel_BINARY_DIR}/OpenBabel3_EXPORTS.cmake")   

2. The package config section::

          configure_file(${openbabel_SOURCE_DIR}/openbabel-3.pc.cmake
            ${openbabel_BINARY_DIR}/openbabel-3.pc @ONLY)
          install(FILES ${openbabel_BINARY_DIR}/openbabel-3.pc
            DESTINATION ${LIB_INSTALL_DIR}/pkgconfig)

  and::

          git mv openbabel-2.pc.cmake openbabel-3.pc.cmake

3. dlhandler_win32.cpp::

    handle = GetModuleHandle("openbabel-3.dll");//CMake/VC++2008 build will use this.

4. src/CMakeLists.txt::

  # Set the output name to openbabel-3 (preserves old behaviour). Is this desired?
  set_target_properties(openbabel PROPERTIES
    OUTPUT_NAME openbabel-3)
  remove_definitions(_UNICODE UNICODE)
   
5. Then there's a whole bunch of lines in CMakeLists.txt starting with::

     # Create the Config and ConfigVersion files in the build directory, useful to
  
   and the associated files::

        git mv OpenBabel2Config.cmake.in OpenBabel3Config.cmake.in
        git mv OpenBabel2ConfigVersion.cmake.in OpenBabel3ConfigVersion.cmake.in
```


ghutchis:
Thanks, this has irritated me since 2.0 was released.

PR #2043, opened 2019-09-23
Bump cmake_minimum_required to 3.1

e-kwsm:
`CMAKE_CXX_STANDARD` is currently used

https://github.com/openbabel/openbabel/blob/41730e2d82d01eb52cc5a58ed60000c533ee8ce9/CMakeLists.txt#L474

But it was introduced in [CMake 3.1](https://cmake.org/cmake/help/v3.1/release/3.1.0.html#properties).


ghutchis:
I have a feeling I'll regret this before a major release, but I suspect this is okay.

PR #2044, opened 2019-09-23
Replace NULL with nullptr (C++11)

e-kwsm:


e-kwsm:
I think Visual Studio 12.0 [already supports `nullptr`](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015?view=vs-2019#whats-new-for-c-in-visual-studio-2010) but AppVeyor fails :confused: 

Actually lambda works (#2001).

ghutchis:
This looks like it's more comprehensive than #2208  - would you be willing to work with @yurivict and look for any remaining NULL in the codebase? I checked your commit here and it still has a bunch of them.

```
/usr/ports/science/openbabel/work/openbabel-openbabel-3-1-0/src/math/align.cpp:448:14: error: C-style cast from 'nullptr_t' to 'double' is not allowed
      return (double) NULL;
```

This one is easy - it should be `return nan;` instead.

ghutchis:
Incidentally, I'd like to do more `clang-tidy` - perhaps we can start a discussion on the mailing list about which checks to start? Probably best to do one fix per pull request, starting with the various `modernize` checks.

yurivict:
NULL-related C++ issues are gone with these patches.

PR #2049, opened 2019-09-30
Remove smart matching for rigid fragment

n-yoshikawa:
In the current implementation, when a rigid fragment is included in other fragments, it caused strange bonds. This pull request avoids this problem. The performance improved slightly judging by the Platinum dataset evaluation.

### DIFF
--- a/src/builder.cpp
+++ b/src/builder.cpp
@@ -46,6 +46,7 @@ GNU General Public License for more details.
  * - practically speaking, this means they are ordered by the number of atoms
  * */
 
+#include <iterator>
 using namespace std;
 
 namespace OpenBabel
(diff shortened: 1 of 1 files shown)


ghutchis:
Out of curiosity, can you post the Platinum evaluation metrics?

n-yoshikawa:
I added an evaluation of execution time.

This pull request:
```
Time:		 92.61 s
RMSD:		 1.7294983379553486
Bond error:	 0.044128213478314485
Angle error:	 2.34070914600812
Torsion error:	 43.25837938247694
TFD:		 0.2504212088912883
Stereo correct:	 94.34916446789798
```
cf. current master:
```
Time:		 54.10 s
RMSD:		 1.7460046846605375
Bond error:	 0.04876801172897926
Angle error:	 2.493798349968771
Torsion error:	 44.014286454398786
TFD:		 0.26552873159881324
Stereo correct:	 93.84344766930519
```
I expected this pull request would shorten the execution time, because only canonical order is used instead of searching SMARTS pattern, but actually it almost doubled. I'm looking for the reason...

PR #2052, opened 2019-10-06
Mark 3.0 release

ghutchis:
Need to check what else might need to change - certainly NEWS file needs updating.


baoilleach:
Just noticed the Sponsor link, which indicates that KitWare have given $500! News to me. Thanks KitWare! According to the sponsor link, this merits a mention in the README. I would add it to the release notes too.

ghutchis:
That comes thanks to Google Summer of Code, which contributes $500 to each org. We should have another $500 coming at some point.

PR #2058, opened 2019-10-10
Fix import of openbabel

e-kwsm:


ghutchis:
Probably better stated as "fix python documentation and examples. 😄 "

Issue #2015, opened 2019-08-16
Use coordgen library for 2D depiction


ghutchis:
- [ ] I believe this to be a bug with Open Babel
- [X] This is a feature request

Schrodinger has contributed their [coordgen](https://github.com/schrodinger/coordgenlibs) library to RDKit for generating 2D depictions under the BSD license.

By all accounts, it produces high-quality depictions and it would be great to adopt a community library for depictions.

ghutchis:
I'd probably mark this for a 3.1 release. Thoughts @baoilleach ?

ghutchis:
I think the main plan would be to add `coordgen` as an optional compile and when available, it's used in preference to the `mcdlutil.cpp` code.

The main hook is in `ops/gen2d.cpp` right now, which simply calls `generateDiagram()`

lorton:
@ghutchis - just in case it's not obvious, coordgen doesn't have a renderer in it (RDKit already had 2 of those, so we decided not to add to the party).  OpenBabel would probably still have to use whateverer renderer it's using, but coordgen could give 2D coordinates that would render better across small molecule ligands, materials compounds and macrocycles.

ghutchis:
Yes, we also have several render options, including SVG, PNG, ASCII.. But we'd definitely like better 2D coordinates, which is why `coordgen` is appealing. There are several open issues and `mcdl` isn't really maintained anymore (e.g., adding new templates is difficult).

lorton:
OK - sounds good.  So from what I can tell we could add coordgen as an optional dependency (probably turned on by default if Boost is present), and in that case we'd call into coordgen to populate coordinates instead of calling into generateDiagram?  Something like

```
#ifdef COORDGEN
generateDiagramCoordgen(...);  // New OB function to be written wrapping coordgen
#else 
generateDiagram(...);
#endif 
```

Are we happy with always using coordgen when it's built in, and building in by default if Boost is available?  If so then I could probably get a PR up for this relatively quickly.

ghutchis:
Perfect - both the pseudo-code and building by default if Boost is available are fine. The Linux packing folks would probably want a Cmake option USE_SYSTEM_COORDGEN to use a pre-installed library and includes (e.g., USE_SYSTEM_RAPIDJSON as an example)

lorton:
[...]

PR #2064, opened 2019-10-21
Find or build maeparser & coordgen libraries

ricrogz:
This is meant as an improvement: with these changes, CMake first checks if there is a build of Schrödinger's maeparser and coordgen libraries that can be used in openbabel's build. If there isn't, then, this patch downloads the sources and builds them.

This means that it addresses #2015.

Some notes:
- CMake will, by default, look in standard library and include directories for the built library and the required headers. Other directories can be specified for the lookup by using "maeparser_DIR" and "coordgen_DIR".
- By default, the "master" branch of maeparser and coordgen will be downloaded from GitHub. This can be changed using the "MAEPARSER_VERSION" and "COORDGEN_VERSION" variables, pointing them to specific tags, branches or commits.
- If sources are already present at the directory where CMake would put them, the download step will be skipped, and the available sources will be used.
- Both maeparser and coordgen test executables will not be built. In case they are, ctest won't be able to find them, since they will be put in the "bin" directory, together with the rest of the openbabel executables.
- Boost threads need to be disabled in case of a static build, or the build will fail (I haven't investigated this too much, but probably the Boost libraries are not built using --static-libgcc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -498,7 +498,7 @@ if(WITH_MAEPARSER)
 
     else()
 
-      set(MAEPARSER_VERSION "v1.2.2" CACHE STRING "Maeparser fallback version to download")
+      set(MAEPARSER_VERSION "master" CACHE STRING "Maeparser fallback version to download")
 
       set(MAEPARSER_DIR "${openbabel_SOURCE_DIR}/external/maeparser-${MAEPARSER_VERSION}")
 
@@ -517,10 +517,7 @@ if(WITH_MAEPARSER)
           execute_process(COMMAND ${CMAKE_COMMAND} -E tar zxf "maeparser-${MAEPARSER_VERSION}.tar.gz"
               WORKING_DIRECTORY "${MAEPARSER_DIR}")
 
-          message(STATUS "${MAEPARSER_DIR}")
-
           find_path(MAEPARSER_UNPACK_DIR "CMakeLists.txt" PATH "${MAEPARSER_DIR}/*" NO_DEFAULT_PATH)
-          message(STATUS ${MAEPARSER_UNPACK_DIR})
 
           if(MAEPARSER_UNPACK_DIR)
             file(RENAME "${MAEPARSER_UNPACK_DIR}" "${MAEPARSER_DIR}/maeparser")
@@ -558,7 +555,7 @@ if(WITH_COORDGEN)
 
     else()
 
-      set(COORDGEN_VERSION "v1.3.2" CACHE STRING "Coordgen fallback version to download")
+      set(COORDGEN_VERSION "master" CACHE STRING "Coordgen fallback version to download")
 
       set(COORDGEN_DIR "${openbabel_SOURCE_DIR}/external/coordgen-${COORDGEN_VERSION}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -482,31 +482,123 @@ find_package(Boost)
 if(Boost_FOUND AND BUILD_SHARED)
     include_directories(${Boost_INCLUDE_DIRS} ${Boost_INCLUDE_DIR})
     option(WITH_MAEPARSER "Build Maestro support" ON)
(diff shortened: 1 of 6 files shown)


ricrogz:
I included the coordgen part because I thought it might be helpful for implementing it and trying it out. Also, I wanted to plan a bit ahead, to check that maeparser finding/building was incorporated in a way that was compatible with adding in coordgen.

I'm ok with both options, although I think it might be best if I removed the coordgen part and the maeparser one was merged, even if just to update it to a more recent version.

Maybe @lorton has a preference about this?

lorton:
@ghutchis - I don't think there's a rush here, but once coordgen is built/linked I'll be able to proceed with the case in adding the ability to optionally use coordgen to generate coordinates.  We're in a bit of a chicken and egg because I'm hesitant to try to contribute code to use coordgen until it's building cleanly in openbabel, and openbabel doesn't have priority to build cleanly until coordgen is being used ;)

CMakeLists.txt
```diff
@@ -482,33 +482,125 @@ find_package(Boost)
...
+
+    find_package(maeparser MODULE QUIET)
+
+    if (maeparser_FOUND)
+
+      message(STATUS "Maestro formats will be supported. Using MaeParser libraries at ${maeparser_LIBRARIES}.")
+
+    else()
+
+      set(MAEPARSER_VERSION "master")
```
ghutchis: I'm a bit hesitant at that. I don't mind "let's use master for OB master".. is the API for maeparser fixed? After all, if maeparser changes, does the code here in OB need an update too?

CMakeLists.txt
```diff
@@ -482,33 +482,125 @@ find_package(Boost)
...
+
+    find_package(coordgen MODULE QUIET)
+
+    if (coordgen_FOUND)
+
+      message(STATUS "Coordinate generation with Coordgen will be supported. Using Coordgen libraries at ${coordgen_LIBRARIES}.")
+
+    else()
+
+      set(COORDGEN_VERSION "master")
```
ghutchis: Ditto.

CMakeLists.txt
ricrogz: That can be changed, of course. In fact, I just noticed a couple of things I forgot about when I wrote that. I'll update shortly.

ricrogz:
In this update:

- Swapped the default "master" versions with the last released versions of maeparser and coordgen.
- Made the version strings cached, so that the versions can be overriden from the `cmake` compile line (`-DMAEPARSER_VERSION=...`).
- GitHub does one kind of dumb thing with the releases: while non-release tags like "master" are unpacked into directories like "[project]-[tag]" (e.g. "maeparser-master"), the release tags are named like "v[x].[y].[z]", but when they are unpacked, the directories are called "project-[x].[y].[z]" (dropping the "v"), so we need to account for that when in the path to the sources.

ricrogz:
Oh, I see why I used maeparser's master: v1.2.2 (the latest) did not yet include the flag to disable the unit tests... Other than that, it is fine.

ghutchis:
@ricrogz - I can't merge something that fails tests. If `master` is needed for that, no problem. I just want to make sure that we can freeze versions for e.g openbabel-3.1.0

I'm aware of the annoying GitHub tar/zip variance. Happy to work on that.

ricrogz:
> @ricrogz - I can't merge something that fails tests. If `master` is needed for that, no problem. I just want to make sure that we can freeze versions for e.g openbabel-3.1.0

Yeah, sorry, I forgot that the switch to disable the test was only in maeparser's master. I'll restore the "master" versions for now.

I didn't really answer your question: the APIs to maeparser and coordgen are rather stable (we don't make many changes that require updates to calling code, and I'm happy to assist when that happens), so it should be ok to keep it set to "master", and change it to whatever version is the current one when you prepare for a release.

> I'm aware of the annoying GitHub tar/zip variance. Happy to work on that.

I gave it a try already, using `find_path` to search for the directory in which maeparser/coordgen's CMakeLists.txt lives. That seemed to work fine in the CI builds.

PR #2072, opened 2019-11-04
Fixed compiler warnings in libinchi

dspoel:
Not sure whether I pushed this before but it would be great to get rid of as many compiler warnings as possible.


baoilleach:
I want the warnings to disappear too, but patches should be applied upstream. This isn't even the latest version of the library.

PR #2074, opened 2019-11-05
Fixed warnings in libinchi

dspoel:
The stack array kLeast_rho_fix is initiated as an array with
MAX_LAYERS entries. It was used as a pointer in statements to
test whether the pointer is NULL. The clang compiler did not
like this and therefore an alternative test for MAX_LAYERS>0
is used.


ghutchis:
As much as I'd like to fix inchi errors, they made it quite clear they don't want people messing with their code 🙄  I don't believe they've accepted any patches to 1.0.x, but hopefully there will be an open period to comment / question 2.0.

PR #2075, opened 2019-11-06
Comment out unused parameters

e-kwsm:


ghutchis:
Sorry, I'm not going to merge anything C++17 anytime soon. C++11 is okay, though. We need to be intentionally very conservative with support.

e-kwsm:
The attribute is not effective unless `-std=c++17` is supplied. Actually there are many unused variables.

correaa:
> Sorry, I'm not going to merge anything C++17 anytime soon. C++11 is okay, though. We need to be intentionally very conservative with support.

Good call.

I think the fallback case should be to use nothing. (evaluate `UNUSED(x)` to `` --nothing--).  https://github.com/openbabel/openbabel/issues/2376

Besides `[[maybe_unused]]` is only for C++17, the keyword is not for that™.
`[[maybe_unused]]` is mainly for generic code or macros, `if constexpr`, etc.

An unused variable shouldn't be named in the first place.
At most have a commented name `f(double /*unused x*/)` (Linters recommend that).
There is nothing wrong with not naming the unusued variable, there is no need to make acrobatics with `unused` attributes.

e-kwsm:
The original motivation is to replace `UNUSED` macro with `[[maybe_unused]]` attribute, so this does not fix all of the warning on unused parameters.

e-kwsm:
Now the unused parameters are commented out as far as clang-tidy does.

PR #2079, opened 2019-11-08
Support #0 in SMARTS so that asterisks can be matched

baoilleach:
The SMARTS parser did not support #0 to indicate an asterisk. This PR adds support...by removing two lines from the source code. :-)


ghutchis:
I'm not 100% sure why anyone would want this, but it certainly doesn't make anything worse.

baoilleach:
I needed it yesterday. I chopped side chains off residues in peptides, and
then I want to search through them matching the attachment point.

PR #2084, opened 2019-11-15
Wiswesser Line Notation reader

baoilleach:
A contribution from @rogersayle of NextMove Software. Note that the reader does not offer full support for all aspects of the language but does cover a lot of ground. Thanks also to @richardjgowers for a bug fix.

The code is structured as two files to allow easy updating by @rogersayle from the original codebase.

The following poster has more info: https://www.nextmovesoftware.com/posters/Sayle_WisswesserLineNotation_BioIT_201904.pdf. Actually I should probably add info from this poster into the docstring. Marking PR as work in progress...


baoilleach:
Removed "WIP" designation. Now ready for merging or comments...

Issue #2095, opened 2019-11-25
SVG generation with --genalias causes segfault


au1985:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: Openbabel 3.0.0 built from git today
Operating system and version: MacOS X 10.14.5

## Expected Behavior
I am trying to convert a molecule with an alias from .smi into an .svg file. Without aliases, everything works fine. I was able to convert the molecules with some default aliases, but sometimes it broke and I got segfault. I created a minimalistic example to reproduce the behavior, see below.

## Actual Behavior
$ ./obabel test.smi -Otest.svg --genalias
Segmentation fault: 11

## Steps to Reproduce
0. Build openbabel
1. Create test.smi file with smiles for thioanisole
$ echo 'c1ccccc1SC' > test.smi
2. --genalias uses superatom.txt file, so modify it to make it minimalistic. I left only a thiomethyl group in it. In my setup, superatom.txt is here:
echo 'SMe   MeS   SC   red' >  /usr/local/share/openbabel/3.0.0/superatom.txt 
3. 
$ ./obabel test.smi -Otest.svg --genalias
Segmentation fault: 11

PR #2102, opened 2019-11-29
Fix issue #2095, caused by incorrect application of protonated SMARTS…

au1985:
alias.cpp contains code that replaces some groups defined in superatom.txt file, with the corresponding aliases, like C(=O)C with Ac. The aliases from superatom.txt are always considered protonated, and implicit hydrogens are added to them to avoid replacing fragments in the middle of the molecule, so e.g. C(=O)CC should not be replaced with Ac. However, the molecule of interest can lack implicit hydrogens. In case of non-protonated molecules, out of range error can happen here:

 `int id  =(pmol->GetAtom(idx))->GetId();`

when it tries to find and remove all atoms found in alias including hydrogens lacking in the molecule. So idx can easily exceed the number of atoms in pmol. It causes out of bounds error and segfault as shown in my minimal example in issue description. Guess that incorrect behavior rather than segfault can also be a result.
The fix creates a protonated copy of the molecule to find numbers of all atoms that should be stripped and replaced with aliases. Hydrogens that have numbers above number of atoms in original molecule are ignored. This should ensure that hydrogens present in the original molecule are retained.

### DIFF
--- a/src/alias.cpp
+++ b/src/alias.cpp
@@ -377,9 +377,14 @@ bool AliasData::AddAliases(OBMol* pmol)
           }
           else
           {
-            AllExAtoms.insert(idx);
-            int id  =(pmol->GetAtom(idx))->GetId();
-            ad->AddExpandedAtom(id);
+            OBAtom* a = pmol->GetAtom(idx);
+            // atom might not be present in original molecule, because SMARTS are all with hydrogens added
+            // original molecule might lack them
+            if (a != NULL) {
+              AllExAtoms.insert(idx);
+              int id  = a->GetId();
+              ad->AddExpandedAtom(id);
+            }
           }
         }
         if(ad)
--- a/src/alias.cpp
+++ b/src/alias.cpp
@@ -344,9 +344,13 @@ bool AliasData::AddAliases(OBMol* pmol)
     LoadFile(smtable);
   set<int> AllExAtoms;
   SmartsTable::iterator iter;
+  // create protonated copy of the molecule for mathing against SMARTS that are all protonated
+  OBMol* protonated = new OBMol(*pmol);
+  unsigned int pmol_atom_count = pmol->NumAtoms();
+  protonated->AddHydrogens();
   for(iter=smtable.begin();iter!=smtable.end();++iter)
   {
-    if((*iter).second->Match(*pmol))
+    if((*iter).second->Match(*protonated))
     {
       vector<std::vector<int> > mlist = (*iter).second->GetUMapList();
       for(unsigned imatch=0;imatch<mlist.size();++imatch) //each match
@@ -357,6 +361,13 @@ bool AliasData::AddAliases(OBMol* pmol)
         for(unsigned iatom=1; iatom<mlist[imatch].size();++iatom)//each atom in match
         {
           int idx = mlist[imatch][iatom];
+          // assume that all atoms added during protonation are added 'on top' of the non-protonated molecule,
+          // so they have ids above pmol_atom_count
+          if (idx > pmol_atom_count) {
+            assert(protonated->GetAtom(idx)->GetAtomicNum() == OBElements::Hydrogen);
+            continue;
+          }
+
           if(AllExAtoms.count(idx))
           {
             //atom already appears in an alias so abandon this (smaller) alias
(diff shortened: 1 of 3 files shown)


ghutchis:
Thanks. The fix makes sense - can you add a unit test. Otherwise, I'm sure this will creep back again.

au1985:
> Thanks. The fix makes sense - can you add a unit test. Otherwise, I'm sure this will creep back again.

Will definitely do that. I am not familiar with cmake yet, trying to figure out how to do it. BTW is there any documentation how to write tests? README in /test contains a todo list...

baoilleach:
Does this patch always copy the molecule? If so, please think about an alternative way to address the problem as this would be is a major slowdown for all users of Open Babel.

au1985:
> Does this patch always copy the molecule? If so, please think about an alternative way to address the problem as this would be is a major slowdown.

I thought about it already. But will matching non-protonated molecule against protonated SMARTS at all give correct results every time??

baoilleach:
I'll have to look into the problem in more detail to see whether there is an alternative, but copying and addng hydrogens will be a performance hit. I've seen this elsewhere - we use this approach to handle matching against SMARTS with explicit hydrogens (I think I might even be the author of this) - these patterns are an order of magnitude slower to match (according to a SMILES benchmark from @johnmay).

If you can't see any alternative, is it okay if I take a look?

au1985:
> I'll have to look into the problem in more detail to see whether there is an alternative, but copying and addng hydrogens will be a performance hit. I've seen this elsewhere - we use this approach to handle matching against SMARTS with explicit hydrogens (I think I might even be the author of this) - these patterns are an order of magnitude slower to match (according to a SMILES benchmark from @johnmay).
> 
> If you can't see any alternative, is it okay if I take a look?

Sure it's OK - please feel free to look. I am an old user of Openbabel, but this is my first participation in the project. I already have code that solves the issue without copying (should I post another pull request for it, or how should it be handled to show it to you? It is very minimalistic change, two lines from master branch). After a match (it seems to work in the cases that I tested), I just check whether atom idx in match is not higher than count of atoms in the starting molecule. Only H atoms from SMARTS have higher idx, so I ignore them and it seems to work. I am however unsure it will work all the time.

johnmay:
IIRC in [CDK Depict](https://www.simolecule.com/cdkdepict/depict.html) I use canonical SMILES rather than substructure matching to assign the labels. Like matched pairs you can make a cut and then lookup a fragment... that way you can handle it there, have a much larger dictionary of terms, and also make it order independent.

Maybe there's some extra code that modifies the patterns if you swapped CO2H and CO2- in that it would break as one is a substructure of the other...

```
*C(=O)O CO2H
*C(=O)[O-] CO2-
```

Personally I would say the fix is not to automatically change the molecule but rather enforce that hydrogens have to be added before calling this method. i.e. a better error than "out of range".

au1985:
@johnmay unfortunately I do not know how SMARTS matching is implemented - need this to discuss efficiency vs SMILES matching. It is a deeper layer of OpenBabel that makes this job. I guess that some extra work is needed in your approach to cope with stereochemistry and isotopes.

Here I wrote a test for alias feature, and also removed copy construction of OBMol in AliasData::AddAliases, just checking if atom exists. Appreciate team members to take a look when you have time.

baoilleach:
Sounds like you're making progress, but perhaps you've forgotten to check in some changes. I still see the copy in the diff of alias.cpp (above, under Files Changes, line 348).

au1985:
@baoilleach noted, thank you. Sorry that the lines reappeared after my lame usage of git, tried to rollback a wrong commit. Seems like have fixed it now.

baoilleach:
Thanks for this. Merging now...

baoilleach:
Just a note to say that I'm putting together the release notes for 3.0.1. I usually include the github name and people's real names. I don't have yours - contact me asap (my username at gmail.com) if you wish me to include it.

au1985:
Hi, thank you - my name is Alex Ustinov. Best regards,Alex  28.04.2020, 11:09, "baoilleach" <notifications@github.com>:
Just a note to say that I'm putting together the release notes for 3.0.1. I usually include the github name and people's real names. I don't have yours - contact me asap (my username at gmail.com) if you wish me to include it.

PR #2109, opened 2020-01-02
Add the --neutralize operation to convert charged atoms to neutral

baoilleach:
Together with the ``-r`` option for "strip all but the largest component", neutralize is often an essential first step for standardising a dataset of molecules (especially as provided by vendors).

Here's the help text from ```obabel -L neutralize``` (some additional background at https://baoilleach.blogspot.com/2019/12/no-charge-simple-approach-to.html):

```
One of the ops
neutralize    Neutralize +1 and -1 charges

Note: this method accepts an optional argument 'changed' which is
described below.

This method uses a simple procedure to generate the neutral form of a
molecule. It does not attempt to balance charges but simply to convert
all atoms with a +1 or -1 charge to neutral by addition or subtraction
of H+.

To a first approximation the procedure is simply to identify all atoms with
either a +1 or -1 charge, set their charges to zero and adjust their
hydrogen counts by -1 or +1 (i.e. we are adding/removing H+). The first
minor issue is that +1 charged atoms must have a hydrogen, or otherwise
we can't remove H+. The second issue is that we must avoid altering
charge-separated representations of multiple bonds, such as nitro which
is often represented as [N+](=O)[O-]. It does this by checking whether a
neighbor atom has a negative charge (for the +1 atoms) or a positive
charge (for the -1 atoms).

If specified, the optional argument 'changed' causes the method to return
True if the molecule was changed by the neutralize operation, and False
otherwise. This is mostly useful if using the API, but for command-line
usage (e.g. via obabel) this filters out molecules that are unchanged
by the operation and only retains those that are changed.
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -79,6 +79,7 @@ set(ops
   ops/gen3d.cpp
   ops/largest.cpp
   ops/loader.cpp
+  ops/neutralize.cpp
   ops/opsplit.cpp
   ops/optransform.cpp
   ops/partialcharges.cpp
--- a/src/ops/neutralize.cpp
+++ b/src/ops/neutralize.cpp
@@ -1,0 +1,127 @@
+/**********************************************************************
+neutralize.cpp - The option --neutralize neutralizes charged atoms
+
+Copyright (C) 2020 by Noel O'Boyle
+
+This file is part of the Open Babel project.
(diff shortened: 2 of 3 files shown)


baoilleach:
Just realised that I should probably add a call to suppress hydrogen's at the start.

ghutchis:
Looks great, thanks.

PR #2114, opened 2020-01-08
Fix issue #1794, UFF atom typing for deuterium

ghutchis:
- data/UFF.prm: Add line for atom type 'D'.
- src/forcefields/forcefielduff.cpp: Add checks to prevent segfaults.
- test/regressiontest.cpp: Add test case.


ghutchis:
Thanks to @timvdm for the fix - just created the pull request.

Issue #2111, opened 2020-01-06
Segfault on SMI to InChI conversion when using "-d"


baoilleach:
With OB 3.0:
```
$ obabel -:"[H][C@@H](I)F" -d -oinchi
Segmentation fault (core dumped)
```
Presumably some interplay between the explicit and implicit H messes up the stereo object...?

baoilleach:
Similarly for cis/trans stereo:
```
obabel -:"[H]/N=C/1\N(CCS1)CC(c2ccccc2)O" -d -oinchi
```

baoilleach:
And another tet stereo case, but which doesn't involve an implicit H: ```[C@@H]12C[C@H](OCC3=CC=CC=C3)[C@@H](COCC3=CC=CC=C3)[C@]1([H])O2```.

timvdm:
The stereo objects are correctly updated (with OBStereo::ImplicitRef) when the hydrogens are removed. However, the InChI format did not handle this correctly.

The first 2 examples are actually a special case. The molecules are not stereogenic because there are two implicit neighbors. These should be (explicitly) ignored, similarly how C[C@](C)(C)C would be handled.

For tetrahedral stereochemistry the inchi format could only handle implicit refs if these were the first or second atom in the smiles string. This was easily extended to the general case.

Fix with regression tests (pull request created): https://github.com/timvdm/openbabel/tree/issue2111

baoilleach:
Thanks for looking into this Tim. My own feeling is that there are two separate issues here:

1. Can an atom with a hydrogen and a lone pair be stereogenic? This is related to stereocenter perception. Let me ping @johnmay on this. I'm pretty sure there are cases of this related to S (maybe not N), but I think John can clear this up.
2. Should RemoveHydrogens() suppress hydrogens involved in cis/trans stereo? I think the answer to this is no. One of the invariants of the toolkit should be that suppressing and expanding hydrogens should not change any information.

Note that (1) and (2) are separate issues. I've changed the SMILES reader so that stereo perception is optional ("-aS"). By default, stereo is just read directly from the SMILES (same behaviour as OEChem and CDK, at least). This improves performance when reading SMILES, and preserves stereo present in the input even where our stereoperception would remove it:
```
$ obabel -:"C[C@](C)(C)C" -osmi
C[C@](C)(C)C	
```

johnmay:
[...]

PR #2115, opened 2020-01-09
Fix segfault on SMI to InChI conversion when using "-d"

timvdm:
See https://github.com/openbabel/openbabel/issues/2111 for details.


baoilleach:
Hi Tim, good to hear from you again. 

Regarding the UFF typing, can you coordinate with Geoff over at https://github.com/openbabel/openbabel/pull/2114 as he already has sent in a PR for this.

Regarding the #2111, I think this needs some more discussion. Let's do that over there.

baoilleach:
Fixes #2111.

PR #2149, opened 2020-03-31
OBBuilder: Add error message for ring/rigid fragments with all zero coords.

timvdm:


ghutchis:
I think I'd two minor pieces:
- a check in GetNewBondVector() and/or OBBuilder::Connect() to look for invalid connection vectors
- Something like this for tests: https://gist.github.com/ghutchis/381adeba95b5d9923757c159757857e8

ghutchis:
I'm going to merge this, since the test was in #2150

PR #2153, opened 2020-04-02
Improve tools/obenergy.cpp

e-kwsm:
- check if input file is passed to avoid SEGV
- remove unused variables
- clang-format


e-kwsm:
At present `obenergy -h`, `obenergy -ff UFF` etc causes SEGV.
I've fixed it (show error message and exit).

PR #2154, opened 2020-04-02
Improve tools/obconformer.cpp

e-kwsm:
* Enable to accept forcefield (default: MMFF94)
* remove unused variables
* clang-format

### DIFF
--- a/doc/obconformer.1
+++ b/doc/obconformer.1
@@ -9,6 +9,7 @@
 .Ar #-of-conformers
 .Ar #-of-optimization-steps
 .Ar filename
+.Ar [forcefield]
 .Sh DESCRIPTION
 The obconformer tool can be used as part of a conformational study
 by generating random conformers using a Monte Carlo search. The best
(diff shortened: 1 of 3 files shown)


e-kwsm:
At present it is impossible to change the forcefield from MMFF94, so we cannot perform obconformer on some systems, e.g., borane.

In this PR I simply add forcefield as the last argument—use it if passed, otherwise defaulted to MMFF94.

ghutchis:
I don't use this tool directly, or i would have noticed the issue. Adding an option is a good choice, but the default should be:
- use MMFF94 if possible
- if not, use UFF

if the user actually asks for a specific forcefield, then try it or fail

If you're willing to make that small change, i'll happily accept the patch, thanks.

e-kwsm:
> Adding an option is a good choice, but the default should be:
> - use MMFF94 if possible
> - if not, use UFF
>
> if the user actually asks for a specific forcefield, then try it or fail

Implemented.

ghutchis:
Thanks - definitely improved 😄

PR #2157, opened 2020-04-03
Fix the import of Tkinter in Python 3+

mrakitin:
While following the example at https://open-babel.readthedocs.io/en/latest/UseTheLibrary/PythonInstall.html#test-the-installation, noticed that the `mol.draw()` step fails on importing of `Tkinter` in Python 3.7. The module was renamed to `tkinter` in Python 3+ versions, and the `six` library supports the proper handling of the change: https://six.readthedocs.io/#module-six.moves (thanks to @tacaswell for the reference). I added `six` as the install requirement.

### DIFF
--- a/scripts/python/openbabel/pybel.py
+++ b/scripts/python/openbabel/pybel.py
@@ -51,7 +51,10 @@ else:
     from . import openbabel as ob
     _obfuncs = _obconsts = ob
     try:
-        from six.moves import tkinter as tk
+        if sys.version_info[0] >= 3:
+            import tkinter as tk
+        else:
+            import Tkinter as tk
         from PIL import Image as PIL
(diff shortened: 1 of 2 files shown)


baoilleach:
A patch that fixes this without using six would be great.

mrakitin:
Why is six a problem? It properly takes care of the py2->py3 transition, is available for all platforms on PyPI and conda.

baoilleach:
Nothing wrong with six, per se. We avoid adding dependencies unless
necessary.

mrakitin:
@baoilleach, please check the fix.

baoilleach:
Thanks Maksim. Looks great.

mrakitin:
You are welcome, Noel! Thanks for accepting the fix.

PR #2165, opened 2020-04-08
Initial CMake GitHub Action

ghutchis:
Thanks to Cristian Adam for a template (and blog article)
https://cristianadam.eu/20191222/using-github-actions-with-c-plus-plus-and-cmake/
https://raw.githubusercontent.com/cristianadam/HelloWorld/master/.github/workflows/build_cmake.yml

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -29,7 +29,7 @@ jobs:
             os: windows-latest,
             build_type: "Release", cc: "cl", cxx: "cl",
             environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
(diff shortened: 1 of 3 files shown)


ghutchis:
TODO:
[ ] Use Eigen3
[ ] Use SWIG - maybe better to be a separate build?
[ ] Separate static build -DBUILD_SHARED=OFF -DWITH_INCHI=ON -DWITH_STATIC_INCHI=ON

baoilleach:
It would be great to have a build on the oldest OS we plan to support e.g. CentOS 7.x (we can ask around for input on this). Once we have this in place we can stop having arguments/discussions over whether we support a particular syntax - if it builds with system GCC on this old system, then it's fine. It looks like this needs docker to be set up somehow, so maybe not an immediate priority.

ghutchis:
@baoilleach - I've seen two rpm builders - the best looks like this one:
https://github.com/marketplace/actions/rpm-build

We'd want the `.spec` file though.

PR #2173, opened 2020-04-11
Try the Windows build again

ghutchis:


ghutchis:
This version still has problems finding the Eigen headers.

Any help would be much appreciated since I don't have a Windows system.

PR #2175, opened 2020-04-12
Add override keyword

e-kwsm:


ghutchis:
This is worth merging for 3.2 (i.e. making the codebase more modern) but it would be great to add a `clang-tidy` check to the GitHub actions to ensure future code keeps the override.

dspoel:
How about using an enum class for the index? That probably has other consequences upstream though.

PR #2176, opened 2020-04-13
Merge of "Improve OBConversion::WriteString() and WriteFile() initialization" (adalke) #1923

baoilleach:
See discussion over at #1923. This is a merge of work by @adalke.

I needed to rebase, hence the separate PR. I also needed to update some of the tests for changes in the interim.

I've made an attempt to get the tests working on Windows (e.g. normalizing line endings), but I think that some of the tests genuinely fail on that platform, e.g. the binary file handling, but presumably these never worked in the first place.

I haven't made the change described by @fredrikw over at #1923. That test passes for me, so I've avoided making the change - I'm not sure why it's passing now.


baoilleach:
Aha - now it's failing in the tests. So I need to make @fredrikw's change after all. :-)

baoilleach:
Good to go, @ghutchis. The codacy warnings about subprocess refer to debug code that Andrew was using.

And just to add that I commented out the problematic pointcloud test mentioned by @fredrikw in the original PR. It uses a random number generator. *sigh*

baoilleach:
After checking, I don't think so. There are only a few code changes. Is there any github action for api changes?

ghutchis:
The main question is whether headers are changed. I'll see if I can create an action using some of the existing tools, but for now it's human checks. 😄

ghutchis:
Looks like a good improvement, thanks for the work to both @adalke and @baoilleach

PR #2177, opened 2020-04-14
Fix the cif write test (i.e. modified format)

ghutchis:
At the same time, updated comment to current website


ghutchis:
Paging @baoilleach - I guess I messed with line endings?

baoilleach:
Your editor is chopping off terminal whitespace even where it's
significant, e.g. between multilinear quotes. You should set your editor to
highlight this, not automatically delete it.

PR #2183, opened 2020-04-28
Add a build for bindings (eventually will generate source tar)

ghutchis:
This will be a separate build script because eventually this
will generate the *source* with all bindings updated, not a binary
(e.g. to automate releases)


ghutchis:
This will undoubtedly need more dependencies installed (C#, etc.) but I'll tackle those as they come up.

PR #2170, opened 2020-04-10
Fix output format in the cifformat writer

afonari:
Related to https://github.com/openbabel/openbabel/issues/2160.

PR #2184, opened 2020-05-01
Allow minimizing without adding hydrogens - not the default

ghutchis:
There are sometimes needs (e.g. in my research) to minimize
with unusual structures - allow minimize without adding H


ghutchis:
Failure is in the cifwriter test #2170

PR #2187, opened 2020-05-02
Fix build on illumos

wiedi:
This include is required for `alloca()`.

Reference: <https://illumos.org/man/3c/alloca>


ghutchis:
I'm fine including the header, but the guard seems a bit strange. illumos will define __sun?

wiedi:
Yes, as the open-source continuation of OpenSolaris it still identifies as SunOS and defines __sun.
Likely this change will also fix Oracle Solaris, but I haven't tested that.

PR #2190, opened 2020-05-03
Bump version number for 3.1 release

ghutchis:


mbanck:
If this bumps SONAME, maybe it should be openbabel-4.0.0 instead? Just throwing that in here.

ghutchis:
@mbanck - in the 2.x release series, we established that 2.1, 2.2, 2.3, etc. would add API, often changing SONAME because ABI would change. The 3.0 release was backwards-incompatible for API, so I'd only go for 4.0 if we went for backwards incompatible changes.

I'll go look through the SONAME rules - IIRC this adds API but may not change ABI.
Do you know if the ABI checker project is still working?

mbanck:
Don't know about the ABI checker project, sorry.

If it's just an ABI but not an API change, I concur with 3.1.

baoilleach:
I've just come across abicompat on Linux. Available on Ubuntu 18.04 for instance. Haven't had a chance to test it.

What I have tested is whether the new openbabel-3.dll on Windows is a drop-in replacement for the openbabel-3.dll of OB3.0 without causing the old obabel to segfault or otherwise complain. And it works fine.

baoilleach:
I've uploaded files to the pre-release, but I've run out of time today to check them. Tomorrow...

ghutchis:
We may need to tweak slightly for a 3.1.1 release to fix these annoying CMake issues.

baoilleach:
Why not just re-upload a newer 3.1.0? It think there might be more problems downstream if we change the variable names between 3.1.0 and 3.1.1.

PR #2199, opened 2020-05-06
See if converting to DOS / Win line endings fixes some 3D coordgen

ghutchis:
Tim suggested some problems with reading fragments on Windows


ghutchis:
Well *that* didn't work so well...

PR #2204, opened 2020-05-07
Initialize allocated memory to zero.

dkoes:
It is arguable if this is strictly better (you won't see uninitialized
reads in valgrind when debugging memory errors) but seems like good
defensive programming to me.  I don't know of any bugs that are fixed as
a result of not having uninitialized memory, but this was motivated by
another bug involving uninitializd memory and the changes have been
sitting in my fork for a while.

### DIFF
--- a/include/openbabel/typer.h
+++ b/include/openbabel/typer.h
@@ -21,7 +21,7 @@ GNU General Public License for more details.
 #define OB_TYPER_H
 
 #include <openbabel/babelconfig.h>
-#include <openbabel/mol.h>
+#include <openbabel/mol.h>  //needed for THREAD_LOCAL definition
 
 #include <vector>
 #include <string>
--- a/src/formats/MCDLformat.cpp
+++ b/src/formats/MCDLformat.cpp
@@ -237,7 +237,7 @@ private:
     int *  nsum[MAXBONDS];
     bool lflag[MAXFRAGS];
     char strg[MAXFRAGS+1];
-	char * strngs[MAXFRAGS+1];
+    char * strngs[MAXFRAGS+1];
     char tstr[MAXFRAGS+1];
     int  numdups, dupfrag, jump;
     bool jflag;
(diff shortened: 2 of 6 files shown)


baoilleach:
I am not a fan of this. Unless initialising to zero is a requirement of the code logic, all that we're doing here is slowing things down, and papering over potential problems. I would prefer the code to segfault and for us to fix the underlying problem, rather than the code seemingly working fine but actually not. A good example is PR #2038, where the underlying logic problem remains (it's on my TODO) and was only found due to the unitialized memory. 

However, I do think it would be good to run our test suite in an environment that could catch such problems (e.g. address-sanitizer). We actually used to do this back in the day when we used CDash for nightly builds.

ghutchis:
@baoilleach  If someone wants to suggest flags, I'm happy to set up an Ubuntu "sanitize" builder. I'm not 100% up - can you add most/all of the sanitize flags together?

dkoes:
Yes, it is not a clear win, but setting things to zero doesn't automatically mask memory errors - they may still segfault and in this case it become easier to debug since the bug is reproducible.  I was instigated to put these changes in after tracking down a bug that would have been much easier to find if the memory had been zero alloced.

Some of these changes are not slower - for example when the calloc is replacing individually setting each element of the array to zero.

baoilleach:
Can you leave out the one for kekulize.cpp and we can merge the rest? I will come back to this code asap and think about it. I note that there are some conflicts that will need to be addressed before merging. Can you rebase?

Regarding flags, I think you've already done it, but I was just thinking of "-fsanitize=address". I don't think you even need a debug build, but I forget the details.

afonari:
Hopefully not hijacking this thread. But can we add checks for the return value (!= nullptr) of these malloc/calloc calls?

See for example: https://www.viva64.com/en/b/0558/

baoilleach:
Let's keep on topic. Feel free to raise on the mailing list or file a feature request.

dkoes:
> Can you leave out the one for kekulize.cpp and we can merge the rest? I will come back to this code asap and think about it. I note that there are some conflicts that will need to be addressed before merging. Can you rebase?

But kekulize is replacing a malloc+memset with a calloc.  It's no problem to change it back, but the calloc should be faster/cleaner.

baoilleach:
Yes - sorry - I didn't check this as closely as I should. Are you happy to rebase the branch? We've had some problems with this in the past, so best to keep the original branch, make a copy of it (i.e. branch off it), rebase the copy (fixing the conflicts), and force-push.

dkoes:
Sorry for letting this languish, I'm at a loss as to what git magic i'm suppose to work.  It looks like a pretty clean set of commits to me.

ghutchis:
I could not merge to this, so i created #2248 

For future reference, there are two options:
- `git pull origin master` and fix the conflicts manually (the big "resolve conflicts button above will do that")
- create a new branch off of origin master and `git cherry pick 943a1d2` and fix conflicts.

Issue #2197, opened 2020-05-06
Missing Boost dependency in cmake config file


antonio-rojas:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 3.1.0
Operating system and version: Linux x86_64

## Expected Behavior
The openbabel cmake target depends on Boost::system, but that target is not defined unless the Boost cmake config is imported. There should be a find_dependency(Boost) in openbabel cmake config.

ghutchis:
Can you submit a patch / diff? The main problem seems to be with the MAEPARSER bits, but cmake is not my strongest proficiency.

antonio-rojas:
Wait, is boost::system actially used anywhere? This was introduced in 15601ca09353e1c1172cb0c809301a3ddd6908a4, but the only bits that remain from that commit are the cmake changes. Is it a leftover?

baoilleach:
Maybe @dkoes can take a look?

ghutchis:
@antonio-rojas - can you take a look at the fix in #2200

antonio-rojas:
Doesn't work, now the cmake target depends on even more boost libraries
```
set_target_properties(openbabel PROPERTIES
  INTERFACE_LINK_LIBRARIES "/usr/lib/libmaeparser.so;/usr/lib/libcoordgen.so;/usr/lib/libm.so;dl;/usr/lib/libz.so;Boost::program_options;Boost::iostreams;Boost::unit_test_framework;-lpthread"
)
```
I think the `if(Boost_FOUND)` block starting at src/CMakeListst.txt:251 needs to be removed completely, since libopenbabel doesn't use any (dynamic) boost libraries.

dkoes:
libcoordgen and libmaeparser are configured to require boost (although it looks like only maeparser links in a library - boost_iostream).  If I remember correctly, there was some discussion about this at the time and the consensus (but not unanimous opinion) was including boost dependencies for extra features that are not required is okay.

What is the goal? To build without boost?

antonio-rojas:
No, the goal is that the openbabel cmake target doesn't depend on things it doesn't use. libopenbabel itself (not talking about libmaeparser) doesn't use boost, so it shouldn't link to it. And if it did, then the dependency must be declared in the cmake config file, otherwise the boost:: targets can't be resolved.

antonio-rojas:
Just to illustrate what this issue is about, here's a test CMakeLists.txt
```
[...]

PR #2205, opened 2020-05-07
CMake: separate thread import from boost

dkoes:
This in response to #2197

Don't include a dependencies on libboost_system in libopenbabel .  Separate code for pulling in threads from boost check.

I've also fixed the case of the cmake documentation (OpenBabel3 instead of OPENBABEL3).  

I have questions though:

- OpenBabel3_INCLUDE_DIR or OpenBabel3_INCLUDE_DIRS?  There is a mismatch between documentation and code
- Does cmake automatically define OpenBabel3_FOUND?  I don't see where we define it in OpenBabel3Config.cmake.in
- Should the documentation no long describe FindOpenBabel3.cmake since we have the cmake.in file to be installed in the standard place?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -250,9 +250,11 @@ add_library(openbabel ${BUILD_TYPE}
   )
 
 target_link_libraries(openbabel ${libs})
-if(Boost_FOUND AND (BUILD_SHARED OR BUILD_MIXED))
+if(Boost_FOUND)
     include_directories(${Boost_INCLUDE_DIRS})
-    target_link_libraries(openbabel ${Boost_LIBRARIES} )
+endif()
+
+if(BUILD_SHARED OR BUILD_MIXED)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
     find_package(Threads REQUIRED)
     target_link_libraries(openbabel Threads::Threads)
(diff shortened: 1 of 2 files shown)


antonio-rojas:
cmake defines OpenBabel3_FOUND itself if all checks insinde the config file pass, yes

ghutchis:
It should be OpenBabel3_INCLUDE_DIRS because that's the cmake standard. 🤷‍♂️ 

We should describe FindOpenBabel3.cmake because the build process installs *that* using FindOpenBabel3.cmake.in as the template.

ghutchis:
@antonio-rojas - do you think this solves the issue?

dkoes:
The doxygen documentation has an example FindOpenBabel file in it, which is out-of-date:
http://openbabel.org/dev-api/cmake_project.shtml

This is what I suspect should be removed.

antonio-rojas:
> @antonio-rojas - do you think this solves the issue?

As far as boost is concerned, it does. There is still some overlinking going on though: libopenbabel inherits libs from the top-level CMakeLists so it ends up linking to all libraries used somewhere else (like maeparser) - but that's for a different issue I guess.

ghutchis:
@antonio-rojas - great, please feel free to file that one too.

PR #2211, opened 2020-05-08
Fix the compiler warning about large object size

baoilleach:
```
[92/558] Building CXX object src/CMakeFiles/openbabel.dir/mol.cpp.o
In member function ‘void OpenBabel::OBMol::ConnectTheDots()’:
cc1plus: warning: ‘void* __builtin_memset(void*, int, long unsigned int)’: specified size 18446744073709551608 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
```
The change I've made is to guard against a negative value. I think the compiler is maybe being extra cautious. On the other hand, maybe it's a true logic error in the original code - can you take a quick look @ghutchis.

### DIFF
--- a/src/mol.cpp
+++ b/src/mol.cpp
@@ -3167,7 +3167,8 @@ namespace OpenBabel
         _c = NULL;
         for (atom = BeginAtom(i);atom;atom = NextAtom(i))
           atom->ClearCoordPtr();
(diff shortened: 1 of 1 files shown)


ghutchis:
I think it's extra cautious, which is good. There are a variety of warnings that should get turned on and cleaned up before 3.2. I'll merge this for now.

PR #2218, opened 2020-05-14
Fix test failure with Python 3

### LINKED ISSUE #2217: testdistgeom.py fails when run with Python 3

## Environment Information
Open Babel version: 3.1.1
Operating system and version: CentOS 8.1

## Expected Behavior
After a successful build, all tests should pass.

## Actual Behavior

> 99% tests passed, 1 tests failed out of 229
> 
> Total Test time (real) = 137.16 sec
> 
> The following tests FAILED:
>         218 - pytest_distgeom (Failed)
> 

## Steps to Reproduce
 1. Perform an ordinary `cmake` / `make` build.
 2. in the build directory, run `make test`

## Additional Comments
On CentOS 8, the test runner apparently recognizes that the platform Python is Python 3.6, and uses that to run this test.  Running the failing test directly with Python 3 also fails, and this shows that the issue arises from incorrect (for Python 3) character escaping.  The same test passes in the same environment when run with Python 2.7.

jcbollinger:
When run with Python 3.6, test/testdistgeom.py fails with a SyntaxError
about a malformed character escape.  This arises from a failure to escape
literal backslash characters in single-quoted SMILES strings (several
occurrences).  Python 2 accepts this and does the right things with it,
but Python 3 rejects it.

Fixes #2217

### DIFF
--- a/test/testdistgeom.py
+++ b/test/testdistgeom.py
@@ -42,13 +42,13 @@ class TestDistanceGeomStereo(BaseTest):
             'C1CC[C@H]2[C@@H](C1)CCCC2',  # cis-decalin
             'C1CC[C@@H]2[C@@H](C1)CCCC2',  # trans-decalin
             '[C@H]1(NC[C@H]2[C@H]1N2)OC',
-            'Clc1cccc(Cl)c1\C=N\NC(=O)c1cccs1',
-            'O=C1NC(=S)S\C1=C/c1ccco1',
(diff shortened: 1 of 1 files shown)


jcbollinger:
The AppVeyor test failure is not mine.  It appears to indicate an issue in the AppVeyor environment.  Is this a known issue?

ghutchis:
The AppVeyor thing is relatively new, but it's clearly a routine fail - since the Windows builds on GitHub are working,  I'm going to ignore it.

PR #2224, opened 2020-05-15
Fix UB in UFF parameter parsing

### LINKED ISSUE #2223: Undefined behavior in OBForceFieldUFF::ParseParamFile()

## Environment Information
Open Babel version: 3.1.1
Operating system and version: CentOS 8.1

## Expected Behavior
 1. OpenBabel should not attempt out-of-bounds array accesses.
 2. OpenBabel's tests should pass when compiled with GCC 8 and the `-Wp,-D__GLIBCXXX_ASSERTIONS` option

## Actual Behavior
While parsing "param" lines, `OBForceFieldUFF::ParseParamFile()` attempts to access a third byte of each atom type string, even when there are only two (including C string terminator).  The deuterium parameters in the default UFF.prm file provided by OpenBabel trigger this misbehavior in practice.  If symbol `__GLIBCXX_ASSERTIONS` is defined when building OpenBabel with GCC 8 then tests `test_regressions_1794` and `test_ffuff_1` fail.

## Steps to Reproduce
The access attempt is visible by inspection at [forcefields/forcefielduff.cpp:1650](https://github.com/openbabel/openbabel/blob/master/src/forcefields/forcefielduff.cpp#L1650) (it's even commented), and the single-character deuterium atom type can be seen at [data/UFF.prm:152](https://github.com/openbabel/openbabel/blob/master/data/UFF.prm#L152).

The test failures can be reproduced by building with g++ of GCC 8.3, and providing hardening options via the environment, then running the tests.  For example,

```
mkdir build
cd build
export CXXFLAGS="-O2 -Wp,-D_GLIBCXX_ASSERTIONS"
cmake ..
make
make test
```

## Additional Comments
I discovered this while packaging OpenBabel into an RPM.  By default, RedHat's rpm macros supply hardening options to software builds, including defining the `__GLIBCXX_ASSERTIONS` macro.

jcbollinger:
When evaluating atom coordination number, method
OBForceFieldUFF::ParseParamFile() assumed that it may access a third character
of the atom type string, but in fact that produces undefined behavior if the
string is only one character long. A one-character atom type occurs in
practice with the default parameter file (for deuterium).

This change addresses the issue by verifying that the second character of
each atom type string is not a C string terminator as a precondition for
accessing the third. If the second is a string terminator then the same
default behavior is provided as if the type were two characters long.

Fixes #2223

### DIFF
--- a/src/forcefields/forcefielduff.cpp
+++ b/src/forcefields/forcefielduff.cpp
@@ -1647,7 +1647,7 @@ namespace OpenBabel {
         parameter.b = 0; // used for tracking number of angles in 5-coordinate
         parameter.c = 0;
 
-        char coord = vs[1][2]; // 3rd character of atom type
+        char coord = vs[1][1] ? vs[1][2] : '\0'; // 3rd character of atom type, if any
         switch (coord) {
         case '1': // linear
           parameter._ipar.push_back(1);


ghutchis:
Yes, thanks!

PR #2226, opened 2020-05-15
Fix memory leak in spectrophore.cpp

afonari:
See related https://github.com/openbabel/openbabel/pull/2220.


ghutchis:
Looks good, thanks!

PR #2228, opened 2020-05-15
Fixes miscellaneous undefined behaviors

jcbollinger:
Fixes three miscellaneous undefined behaviors:

 1. One instance of using memcpy() with overlapping source and
    destination strings.

 2. One instance of reading and modifying an object without an
    intervening sequence point.

 3. One modification of an object via a pointer to a different,
    unrelated type.


ghutchis:
Great, thanks!

PR #2230, opened 2020-05-16
Add extra include for MSVC2015 support of back_inserter

baoilleach:


ghutchis:
This is fine - can you take a look at the AppVeyor build - it claims it's failing due to:

> LINK : fatal error LNK1104: cannot open file C:\msvc-dependencies-master\libs-common\i386\zlib1.lib

(Or should we just ditch AppVeyor now that we have GitHub windows builds?)

baoilleach:
Should be a trivial fix for Appveyor. I'll sort it out asap.

baoilleach:
If Appveyor keeps chugging along, I see no reason to ditch it - the problem here was that I updated many of the LIBs and DLLs in the msvc-dependencies project in an effort to fix the problem with the Python bindings - this would have broken any Windows build that used these.

I see that it's using MSVC2012 (I think I deliberately used an older one when setting it up) and compiling fine so it's a useful canary. If it starts to fail, we can bump up the compiler version.

PR #2232, opened 2020-05-16
Fix memory leaks in parsmart.cpp and smartstest.cpp

afonari:
See related https://github.com/openbabel/openbabel/pull/2220


ghutchis:
Thanks!

PR #2233, opened 2020-05-16
Add destructor to groupcontrib (fix memory leak)

afonari:
See related https://github.com/openbabel/openbabel/pull/2220


ghutchis:
I'll have to hold off for a bit - adding a declared destructor would mark this for 3.2 (ABI change).

Issue #1934, opened 2019-02-24
smiles to xyz conversion isn't repeatable


yurivict:
With this ```x.smi``` file: ```O=C(O)CCCCC``` I run this command:
```
obabel -i smi x.smi -o xyz -O x1.xyz --gen3d
```

The produces outputs are a little bit different with every run:
```
$ diff x1.xyz x2.xyz 
3,22c3,22
< O          2.80897        0.95953       -0.55185
< C          2.31033       -0.02262       -0.02710
< O          0.97006       -0.19149        0.06313
< C          3.07414       -1.17334        0.58772
< C          4.58550       -0.97549        0.48201
< C          5.34900       -2.14439        1.10655
< C          6.86037       -1.94108        0.99792
< C          7.62678       -3.09920        1.61670
< H          0.70739       -1.02022        0.50592
< H          2.78449       -2.09272        0.06679
< H          2.78449       -1.25134        1.64155
< H          4.87040       -0.04029        0.98021
< H          4.87040       -0.86979       -0.57232
< H          5.06672       -3.07791        0.60475
< H          5.06672       -2.24615        2.16150
< H          7.14793       -1.01060        1.50124
< H          7.14793       -1.84227       -0.05533
< H          7.38728       -3.20534        2.67958
< H          8.70461       -2.93105        1.52686
< H          7.38728       -4.04175        1.11413
---
> O          2.79827        0.94419       -0.78732
> C          2.30225        0.05508       -0.11491
> O          0.96238       -0.10012        0.00253
> C          3.06876       -0.98420        0.67100
> C          4.57978       -0.80310        0.53395
> C          5.34512       -1.85925        1.33266
> C          6.85617       -1.67396        1.19240
> C          7.62428       -2.72048        1.98393
> H          0.70195       -0.85007        0.56965
> H          2.78027       -1.97412        0.30044
> H          2.78035       -0.89719        1.72445
> H          4.86364        0.19878        0.87975
> H          4.86359       -0.86301       -0.52423
> H          5.06357       -2.86003        0.98314
> H          5.06370       -1.79534        2.39085
> H          7.14294       -0.67605        1.54402
> H          7.14291       -1.74056        0.13642
> H          7.38553       -2.65956        3.05050
> H          8.70190       -2.56728        1.86813
> H          7.38557       -3.73020        1.63495
```

The actual molecules are the same, typical ```RMSD``` differences are ```0.00190```, ```0.00102```, ```0.00067```.

[...]

PR #2241, opened 2020-05-25
Reimplement OpenBabel::OBRandom as a wrapper of std::mt19937_64

### LINKED ISSUE #1113: Method to specify seed for ran. gen. in OBConformerSearch
The OBConformerSearch class uses the current time as the seed for the random generator used to create the initial population and the next generation populations. This means that the results cannot be recreated. It would be nice to have a method that allows one to specify the seed thereby enabling easy recreation of earlier results given the same input. The default seed can (and should probably) still be the current time.

Reported by: mkjensen

### LINKED ISSUE #2783: ability to have a consistent 3D orientation from SMILES with `openbabel.pybel.readstring()` and `make3D()`
Using the `SMILES` `readstring` converter in `pybel` gives a different orientation of the molecule each time it is run. 

Is there a way to consistently have the same orientation / configuration? 

this is in order to have reproducible results over multiple types of first-principles calculations. 

given for example: 

```
from openbabel import pybel as pb 
mol = pb.readstring(string='CCCC',format='smi')
mol.make3D(steps=0)
mol.atoms[0].coords # 
```

if you were to put it in a `for` loop: 

```
from openbabel import pybel as pb 
for i in range(10):
    mol = pb.readstring(string='CCCC',format='smi')
    mol.make3D(steps=0)
    print(mol.atoms[0].coords)
```

which gives: 

> 1.07	-0.03	0.06
> 1.05	0.06	-0.07
> 0.95	0.07	-0.05
> 1.01	-0.07	0.07
> 1.08	-0.06	-0.01
> 1.01	0.07	0.07
> 1.00	-0.07	-0.07
> 0.98	0.08	-0.05
> 0.95	0.09	0.01
> 1.06	0.07	0.04

e-kwsm:
Mersenne twister is one of high-quality PRNGs and available on all platforms since C++11.


ghutchis:
Looks like a good idea.

One thing that's been requested to add an OBOp plugin to set the RNG seed (e.g., #1934) - would you be willing to add that too.

baoilleach:
I've said it before but we should be working to remove all uses of random numbers in the library.

ghutchis:
@baoilleach - I don't think that's possible. It's good to minimize them, allow a defined seed to be set, etc. but stochastic methods are critical to several tasks in the library.

ghutchis:
We've generally said that only major version changes (e.g., OB2 => OB3 or OB3 => OB4) would drop functions / methods.

The general concept is great, but can we either keep the existing code for backwards compatibility and *add* a new OBRandomMT class to use in conformer search, etc.? The other alternative would be to replace with the current code but wrap previous methods for backwards compatibility.

Otherwise we have to leave this for a while.

e-kwsm:
Done.

At the moment, i.e. `OB_VERSION < OB_VERSION_CHECK(4, 0, 0)`, the codes still use `OBRandom` and thus are backward compatible (rounding error may occur).
They will be automatically changed to utilize `OBRandomMT` when version 4 is released.
Explanations of `OB_RANDOM_SEED` environment variable in man docs are currently commented out.

alexander-telepov:
@e-kwsm thanks for such opportunity to fix seeds for reproducibility! However after build this pr I still can't repeat results both putting argument in obabel call from bash and after setting environment variable, for example for this CC1=CC(CNC2=CC=C(NC3CCC4=CC=C(C5=NNN=N5)C=C43)C=C2NC2=CC=C(F)N=C2F)=NC(C)=N1 mol. Can you please suggest some workaround?

e-kwsm:
@alexander-telepov You have to build Open Babel with `-DOB_USE_OBRANDOMMT` to enable the feature.

alexander-telepov:
@e-kwsm Thanks for such fast reply!

ghutchis:
@e-kwsm - thanks, I will get to this soon. I know you're updating all your branches - can you hold off a bit. Right now, the tests remain queued because GitHub limits the number of actions (e.g., a lot of branches are waking up right now).

e-kwsm:
> can you hold off a bit. Right now, the tests remain queued because GitHub limits the number of actions (e.g., a lot of branches are waking up right now).

Alright.

ghutchis:
While we're not going to remove the older OBRandom in 3.2, perhaps OBRandomMT should be the default compile option?

src/config.h.cmake
```diff
@@ -203,3 +203,8 @@
     #define TIME_WITH_SYS_TIME 0
   #endif
 #endif
+
+// use OBRandomMT if true, otherwise deprecated OBRandom
+#ifndef OB_USE_OBRANDOMMT
+#define OB_USE_OBRANDOMMT (OB_VERSION >= OB_VERSION_CHECK(4, 0, 0))
```
e-kwsm: > While we're not going to remove the older OBRandom in 3.2, perhaps OBRandomMT should be the default compile option?


```suggestion
#define OB_USE_OBRANDOMMT (OB_VERSION >= OB_VERSION_CHECK(3, 2, 0))
```

will enable the feature by default in 3.2.

ghutchis: I think we should just switch the default in HEAD when this merges so there's some time to test, check fuzz, etc.

ghutchis: I pushed a suggestion - adding a USE_RANDOMMT cmake flag that defaults to ON - but can be compiled off if people prefer the older RNG.

We won't toss the older OBRandom, but 3.2 will bump the ABI so it should work - and allow specifying seeds.

ghutchis: Let me know if you're okay with this or would prefer a different approach.

e-kwsm: LGTM :+1:

ghutchis:
@baoilleach - still some random numbers but now they can be seeded for reproducibility

PR #2243, opened 2020-05-25
Support Javadoc from doxygen for SWIG 4.0 Java bindings

baoilleach:
Requires merging of #2242 first.

I note that only doxygen from header files is included. We should probably standardise on putting documentation there.

I've generated the javadoc from this, and it looks pretty nice.


ghutchis:
We had been putting longer docs into the source files to minimize the time required to compile - anytime a compiler reads the header, it needs to go through all the comments / docs.

I think it's a great idea - later this week I'll see what I can do to get the SWIG4 working on the Github builders. At the moment Ubuntu 20.04 isn't yet available.

PR #2248, opened 2020-06-05
Initialize allocated memory to zero.

ghutchis:
It is arguable if this is strictly better (you won't see uninitialized
reads in valgrind when debugging memory errors) but seems like good
defensive programming to me.  I don't know of any bugs that are fixed as
a result of not having uninitialized memory, but this was motivated by
another bug involving uninitializd memory and the changes have been
sitting in my fork for a while.


ghutchis:
Handled the git conflict from #2204

PR #2254, opened 2020-06-26
Switch SWIG build to Ubuntu 20.04 which should have SWIG4

ghutchis:


ghutchis:
I'm dropping this and will pick up in a new branch.

Issue #1999, opened 2019-06-28
'OBMol' object has no attribute 'Kekulize'


flc:
- [x] I believe this to be a bug with Open Babel

## Environment Information
Open Babel version: 2.4.90 (current master)
Operating system and version: Ubuntu 18.04.2 LTS
Python version: 3.6.8

## Expected Behavior
```
In [1]: import openbabel
In [2]: obmol = openbabel.OBMol()
In [3]: obmol.Kekulize()
Out[3]: True
```

## Actual Behavior
```
In [1]: from openbabel import openbabel
In [2]: obmol = openbabel.OBMol()
In [3]: obmol.Kekulize()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-67b88b46c5fd> in <module>
----> 1 obmol.Kekulize()

AttributeError: 'OBMol' object has no attribute 'Kekulize'
```

baoilleach:
That method no longer exists as the handling of aromaticity and kekulization has been changed.

The good news is that it has been replaced by the global function OBKekulize. The bad news is that the bindings haven't been updated to include the kekulize.h file. If you add "%include <openbabel/kekulize.h>" to an appropriate location in openbabel-python.i, the bindings should be created. I'll try and get in a fix for this in the near future if you don't have the time.

Out of curiousity, why are you interested in this function?

flc:
Hi, thanks for the answer and for the fix/including it in the python bindings in advance!

Some of my codebase uses the kekulize function for dearomatization purposes so this change makes the it backward incompatible.
Is this new OBKekulize class and its usage documented somewhere?

baoilleach:
As documented in
https://raw.githubusercontent.com/openbabel/openbabel/master/include/openbabel/kekulize.h.
This will appear on the web in the API docs once those are regenerated.

Note that the described behaviour is different from the original, which is
why this is an API breakage and part of OB 3.0, rather than OB 2.x.

baoilleach:
Bump. For 3.0.1 at this stage.

PR #2283, opened 2020-09-27
Resolve Issue #1999

dkoes:
Include kekulize.h in bindings so you can run OBKekulize.


dspoel (commented):
Looks good to me.

dspoel:
Except it  would be good with a test or two as well, to show that it works.

Issue #2282, opened 2020-09-26
Segmentation violation in point group code


dspoel:
- [ X] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 3.x
Operating system and version: MacOS

## Expected Behavior
Code works

## Actual Behavior
When compiling the attached file as an OB tool and running
```
pgtest water.sdf
```
The code will crash in routines down below the pointgroup code since the atoms do not have a valid parent pointer.

[pointgroup.zip](https://github.com/openbabel/openbabel/files/5286601/pointgroup.zip)

## Steps to Reproduce
See above.

dspoel:
Offending source code:
```
      OBPointGroup obPG;
      obPG.Setup(&mol);
      printf("Point group according to OpenBabel: %s\n", 
             obPG.IdentifyPointGroup());
```

ghutchis:
Yeah, this code needs to be replaced with `libmsym` which is much, much better:
https://github.com/mcodev31/libmsym

PR #2285, opened 2020-09-28
Fixed crash in obthermo due to pointgroup code.

dspoel:
The pointgroup code has been disabled in obthermo because a
crash  was introduced relating to the link between OBatom and
OBmol.


dspoel:
Resolves #2282

ghutchis:
Yes, it's worth replacing the point group code with msym but I'll hold off on that for a bit until I have more free time.

Thanks!

PR #2295, opened 2020-11-02
Fix ring iterations w/no rings

dkoes:
The constructor was not initializing _ptr to null, and this is what was
used to test the validity of the iterator.  This caused iteration over a
ringless molecule (e.g. empty) to sometimes fail (segfault or infinite
loop).  But not always.


ghutchis:
Thanks for the catch, that's a clear fix.

PR #2299, opened 2020-11-05
Add GetNumHvyAtoms to OBResidue

RMeli:
`OBMol` has both `NumAtoms` and `NumHvyAtoms` while `OBResidue` only has `GetNumAtoms`. This PR adds `GetNumHvyAtoms`  to `OBResidue`.

I am not familiar with OpenBabel's tests. I added simple tests for `GetNumAtoms` and `GetNumHvyAtoms` but advice on how to do this properly is very welcome.


dkoes:
This makes sense to me - a more consistent API and includes a test case.

PR #2315, opened 2021-01-03
Improve the CMake config files

### LINKED ISSUE #2264: CMake is unable to find Open Babel installation

## Environment Information
Open Babel version: 3.1.1
Operating system and version: Debian sid

## Expected Behavior
Open Babel's CMake files help locating Open Babel's installation.

## Actual Behavior
Open Babel's CMake files seem to be pointing to incorrect locations in Debian system. At least `OpenBabel3Config.cmake`:

```
# The OpenBabel3 config file. To get the targets include the exports file.
get_filename_component(OpenBabel3_INSTALL_PREFIX "${OpenBabel3_DIR}/../../.."
  ABSOLUTE)
```
I know very little about CMake, but I cannot find where the value of `${OpenBabel3_DIR}` is set. On Debian systems, `OpenBabel3_INSTALL_PREFIX` most likely is `/usr`, but some other variables have to be adjusted too.

```
set(OpenBabel3_VERSION_MAJOR   "3")
set(OpenBabel3_VERSION_MINOR   "1")
set(OpenBabel3_VERSION_PATCH   "0")
set(OpenBabel3_VERSION         "3.1.0")
```
These variables should not be hardcoded - I am using Open Babel 3.1.1, but their values indicate that this is 3.1.0. The same issue is in `OpenBabel3ConfigVersion.cmake` too.

```
set(OpenBabel3_INCLUDE_DIRS "${OpenBabel3_INSTALL_PREFIX}/include/openbabel3")
```
`OpenBabel3_INCLUDE_DIRS` is correct.

```
set(OpenBabel3_EXPORTS_FILE "${OpenBabel3_INSTALL_PREFIX}/lib/cmake/openbabel3/OpenBabel3_EXPORTS.cmake")
```
On Debian systems this is `/usr/lib/<arch-triplet>/cmake/openbabel3/OpenBabel3_EXPORTS.cmake`, for example, `/usr/lib/x86_64-linux-gnu/cmake/openbabel3/OpenBabel3_EXPORTS.cmake`.

Maybe CMake files could be generated during the build according to the values of `CMAKE_INSTALL_PREFIX` and `CMAKE_INSTALL_LIBDIR` CMake variables?

## Steps to Reproduce
Originally spotted while [building Kalzium with Open Babel support](https://phabricator.kde.org/D29482#674939).

### LINKED ISSUE #2305: Cmake fails to find imported targets maeparser and coordgen

## Environment Information
Open Babel version: 3.1.1
Operating system and version: CentOS 8.2.2004

## Expected Behavior
I expected to be able to build a small C++ application that uses cmake and has openbable as a dependency.  This has worked with previous versions of openbabel in the 2.x.x line.

## Actual Behavior
The packages CMakeLists.txt includes the command 
    find_package(OpenBabel3 REQUIRED)
which successfully locates openbablel's OpenBabel3Config.cmake file but that generates the following error:

  Found package configuration file:

    ~/software/openbabel/3.1.1/lib/cmake/openbabel3/OpenBabel3Config.cmake

  but it set OpenBabel3_FOUND to FALSE so package "OpenBabel3" is considered
  to be NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: maeparser
  coordgen

pinotree:
Use the modern CMake macros to produce config files that properly work (hopefully). This also makes it possible to deprecate the `OpenBabel3_INCLUDE_DIRS` & `OpenBabel3_LIBRARIES` public CMake variables, and use only the `openbabel` imported library.

This fixes #2264.
This fixes #2305.


ghutchis:
Thanks. Considering I have to go through CMake documentation anytime I have to fix some of this, it's greatly appreciated. And I agree it's probably time to move to 3.1 or later.

PR #2324, opened 2021-02-03
Replacing deprecated set-env and set-path commands

yishutu:
For the GitHub workflows, the `set-env` and `set-path` commands have been deprecated and removed.

- https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
- https://github.blog/changelog/2020-11-09-github-actions-removing-set-env-and-add-path-commands-on-november-16/

This PR replaces one `set_env` command in the `build_cmake.yml` according to the GitHub Environment Files documentation.

- https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files

Also, the `peter-evans/create-pull-request` has also been updated to 3.8.0 as well for this issue.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -56,7 +56,7 @@ jobs:
       if: runner.os == 'Windows'
       run: |
         choco install ninja
-        echo "::set-env name=DEPS::msvc-dependencies-master"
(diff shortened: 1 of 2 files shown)


ghutchis:
I'lll handle the SWIG issues in another fix.

PR #2328, opened 2021-02-21
Update built-in data files.

dkoes:
If openbabel can't find the txt datafiles, for many datatypes there is a
built-in version that is compiled in.  This compiled in header was
manually generated by running a script in the data directory.  I
realized these headers haven't been updated in years and are out of sync
with the txt files (the newer versions of which fix several important bugs).  
I've moved generation of these headers into the cmake
system so they are kept in sync and added a few (e.g. spacegroup didn't
have a compiled in version and this was reported as a warning when
people were using smina).


ghutchis:
The breaks seem unrelated. I'd like to get a GitHub action to run these a needed, but I'll do that later.

PR #2338, opened 2021-03-22
Symbols conversion corrected

### LINKED ISSUE #2333: Incorrect conversion of symbols
If a chlorine atom is represented as CL instead of Cl in a file Open Babel, does not properly convert the symbol. This can easily be seen when writing the SMILES out for the file.

## Environment Information
Open Babel version: v3.1.1 conda-forge
Operating system and version: Manjaro 20.2.1, Python 3.8

## Expected Behavior

SMILES: CCl

## Actual Behavior

SMILES: [CH3].*
## Steps to Reproduce

Writing the SMILES string for the molecule below resulted in the [CH3].* SMILES output.
```
from openbabel import pybel

mol = next(pybel.readfile("xyz", "ccl.xyz"))
print(mol.write("smi")
```
5

C         -3.25634        0.95398        0.05312
CL        -1.48936        0.95398        0.05311
H         -3.61336        1.82622        0.60505
H         -3.61337        0.03988        0.53253
H         -3.61338        0.99586       -0.97822

Arraxx:
this issue fixes #2333 
**Now the uppercase characters will also be accepted**
**For Example :**
CL will return the value of Cl and cL will also return the value of cl.

**Code Description :**
Changed the 2 element to lowercase using _tolower()_ function. 

Exception :
For Xe used the _else if_ statement.
Screenshot (162)

### DIFF
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -1459,6 +1459,7 @@ namespace OpenBabel
           break;
         }
         break;
+      }
       return 0;
     }
 
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -174,8 +174,8 @@ namespace OpenBabel
     unsigned int GetAtomicNum(const char* ptr)
     {
       switch (ptr[0]) {
-      case 'A':
-        switch (ptr[1]) {
+      case 'A':                       
+        switch ((char)tolower(ptr[1])) {
         case 'c':
           if (ptr[2] == '\0') { // Ac
             return 89; // Actinium
@@ -219,7 +219,7 @@ namespace OpenBabel
         }
         break;
       case 'B':
-        switch (ptr[1]) {
+        switch ((char)tolower(ptr[1])) {
         case '\0': // B
           return 5; // Boron
         case 'a':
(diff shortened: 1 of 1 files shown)


Arraxx:
Please review @baoilleach @ghutchis

src/elements.cpp
```diff
@@ -1452,7 +1459,6 @@ namespace OpenBabel
           break;
         }
         break;
-      }
```
fredrikw, line 1455: This deletion is probably responsible for the failed tests.

Arraxx: @fredrikw I have made the changes as per your suggestion. Please review.

baoilleach:
The whole point of the switch statement is to be super fast. Adding function calls to tolower() makes no sense. The switch statement is automatically generated and can be adapted to handle every case. But I thought we agree that tolower() would be called by the XYZ format rather than put in here.

ghutchis:
Exactly. The change should not be to elements.cpp but to formats/xyzformat.cpp::ReadMolecule()

```
        int atomicNum = OBElements::GetAtomicNum(vs[0].c_str());
```

That code needs to be smarter. Right now, if it fails (i.e. the atomic number is 0) it assumes it's a number.

```
        if (atomicNum == 0) {
            // check if it's a number - then set it as a number
            // otherwise check for inconsistent capitalization
            // call toupper() on the first character and tolower() on the second character
        }
```

ghutchis:
I'm closing this. Happy to accept changes to `xyzformat.cpp` as indicated above.

PR #2342, opened 2021-03-25
First fuzz targets for integration with oss-fuzz

catenacyber:
cc @ghutchis as discussed in https://github.com/google/oss-fuzz/pull/5492#issuecomment-806180790

The smiles fuzzer quickly finds a buffer overread
The sdf fuzzer quickly finds a memory leak


ghutchis:
The breaks aren't due to this patch, so I'm inclined to merge (and then hunt down the issues).

PR #2346, opened 2021-03-28
use pip to install python interface without pre-installing c++ interface

njzjz:
Use pip to Install from the source code:
Firstly install Eigen3 and swig, then
```sh
git clone https://github.com/openbabel/openbabel
cd openbabel
pip install -v .
```
OpenBabel Python interface will be installed.

So, it is also possible to build wheels. <s>I created a GitHub Action to build automatically.</s> (CI still has some problems but building manually is OK)
If you can upload wheels to the pypi, then users can use `pip install openbabel` to install
OpenBabel directly without pre-installing anything.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -418,10 +418,15 @@ if(UNIX AND BUILD_SHARED)
   if(APPLE)
     set(CMAKE_INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR})
   else()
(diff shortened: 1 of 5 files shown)


njzjz:
I also did some work on v3.1.1. On Linux, you may install openbabel v3.1.1 using the following without any other things
```sh
pip install openbabel-wheel
```
Then you can import openbabel in Python
```py
from openbabel import openbabel
```

ElricleNecro:
Will this be officially supported? Or maintained for each new version of openbabel?

PR #2395, opened 2021-07-01
More robust pdb reading.

dkoes:
Not all istream implementations support tell/seek so use peek when
consuming blank lines.


ghutchis:
LGTM - merging

PR #2456, opened 2022-03-18
Fix search of coordinate system

SebastianWellig:
**When merged this pull request will:**
- Prevent parser from matching ` Dipole orientation:` if no input orientation was printed.
- `obabel H2O_freq.log -O H2O.cml` work with [H2O_freq.log](https://github.com/openbabel/openbabel/files/8307662/H2O_freq.log)


ghutchis:
Thanks - LGTM

PR #2463, opened 2022-04-24
Fix CMake config install location on Windows

### LINKED ISSUE #2462: Wrong install dir for CMake config file on Windows

## Environment Information
Open Babel version: 3.1.1
Operating system and version: Windows 10

## Expected Behavior
OpenBabel3Config.cmake is installed into a suitable location, e.g. `<prefix>/lib/cmake/openbabel3`

## Actual Behavior
OpenBabel3Config.cmake is installed to `<prefix>/bin/openbabel3`, where consumers do not find it

## Steps to Reproduce
Build and install on Windows

The issue is with line https://github.com/openbabel/openbabel/blob/master/CMakeLists.txt#L746
LIB_INSTALL_DIR is 'bin' on Windows, which is undesired here

nicolasfella:
LIB_INSTALL_DIR is 'bin' on Windows, resulting in an unsuitable install location

Instead use the lib dir info from GNUInstallDirs, which gives the appropriate result here

Fixes #2462


mbanck:
Thanks for the pull request, it looks reasonable to me and I checked that the Linux build is not changed by it.

ghutchis:
Thanks - LGTM - the build failures are unrelated.

PR #2468, opened 2022-04-27
test(python): improvements

e-kwsm:
- fix misuse of `unittest.TestCase.assert_`
- improve usages of `unitest.TestCase` methods


ghutchis:
These were all from older Python code - LGTM

PR #2229, opened 2020-05-16
Do not release OpenBabel::OBConformerSearch::m_score

e-kwsm:
Instead callers must release objects.
Fix #2066, fix #2221.

PR #2474, opened 2022-05-11
Do not release OpenBabel::OBConformerSearch::m_filter

szlepermarcin:
related to https://github.com/openbabel/openbabel/pull/2229, current implementation causes segmentation fault after call SetFilter from python


ghutchis:
Thanks - LGTM

PR #2475, opened 2022-05-11
fix verbose stdout write of GetConformers

szlepermarcin:
change std::cout to m_logstream for better output write control


ghutchis:
LGTM - thanks!

PR #2480, opened 2022-05-25
Support for orca4.1 and spectra

ghutchis:
Found this from the Orca forum.. not sure why it was never submitted for a pull request.


ghutchis:
This one needs some significant work. Supporting x-ray spectra is fine, but naming a generic class after Orca is not.

PR #2487, opened 2022-06-09
For now disable SWIG GitHub actions until they can be fixed

ghutchis:
Signed-off-by: Geoff Hutchison <geoff.hutchison@gmail.com>


ghutchis:
Not needed by #2465

PR #2499, opened 2022-06-24
regenerate scripts/R/examples/openbabelR/src/openbabel-R.cpp by SWIG 4.0.2

e-kwsm:


ghutchis:
I think what we need to do is set up a GitHub action that will run the latest version of SWIG to generate all the bindings periodically and submit it as a pull request.

PR #2500, opened 2022-06-24
refactor: use `std::unordered_map` for hash map

e-kwsm:
The checks are no longer necessary as C++11 is now used.


ghutchis:
Thanks - there's a large amount of older code / hacks to peel away. 👍

PR #2509, opened 2022-07-07
update interface pybel.py

zhongxiang117:
There are two minor issues in Python interface `pybel` module `readfile` function:

It has arguments like `readfile(format, filename, opt=None)`, argument `format` is mandatory required, however, it should have the ability to be guessed from the input `filename`, which is also required. Secondly, for `openbabel`, the compressed `gzip` filetype can also be supported in some extents. User may feel confused with which type of “correct” file format should be input if they happen to have a `gzip` file like “**file.sdf.tgz**" or the equivalent things.

To be noted: in order to be maximumly compatible with early version of `pybel`, the sequence of arguments inside are kept consistent.

With the new version of codes, `readfile` can be used to read file like: `readfile(filename=”head.sdf.tgz”)`.

### DIFF
--- a/scripts/python/examples/testpybel.py
+++ b/scripts/python/examples/testpybel.py
@@ -192,7 +192,7 @@ M  END
         self.assertRaises(OSError, self.RFreaderror)
 
     def RFformaterror(self):
-        mol = getattr(self.toolkit.readfile("noel", "head.sdf"), nextmethod)()
+        mol = getattr(self.toolkit.readfile("noel", self.head), nextmethod)()
 
     def testRFformaterror(self):
         """Test that invalid formats raise an error"""
--- a/scripts/python/examples/testpybel.py
+++ b/scripts/python/examples/testpybel.py
@@ -19,9 +19,9 @@ try:
         cinfony = None
     try:
         from openbabel import pybel
-        rdkit = cdk = None
     except ImportError:
         pybel = None
+    rdkit = cdk = None
 except AttributeError:
     from cinfony import cdk
     pybel = rdkit = None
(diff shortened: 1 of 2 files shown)


ghutchis:
This looks really nice, but it's breaking tests in `scripts/python` - which expects a `ValueError` for the invalid format over an `OSError` since the file doesn't exist.

Can you please either update `scripts/python/examples/testpybel.py:
```
    def testRFformaterror(self):
        """Test that invalid formats raise an error"""
        self.assertRaises(ValueError, self.RFformaterror)
```

Or in your code, you should be able to check for an invalid format and *then* do your check for files.

zhongxiang117:
Dear Dr. Hutchison,

Thank you for you pointing out where the conflict of the this pull request.

Actually, this problem is caused by the “priority” of the checking sequence. Let me explain it in detail.

To validate the `file-format`, in testing script `scripts/python/examples/testpybel.py`, there is an intentioned `assert`:

```
LINE 194:  def RFformaterror(self):
LINE 195:      mol = getattr(self.toolkit.readfile("noel", "head.sdf"), nextmethod)()
```

In old `pybel.py`, since the format `noel` is not supported at the very beginning, the `format-error` will be raised firstly, by using returned Boolean variable `formatok`, which is just like assumed. 

However, in new updated codes, the function becomes: `pybel.readfile(format=None, filename=None, opt=None)`, because the argument `filename` is always required, so to be efficient, it is being checked first.

Due to file path of `head.sdf` is `python/examples/head.sdf`, but does not exist at `where-the-ctest-being-executed`, thus the error `IOError` will be raised, yet it is not equivalent with the expected `ValueError`, coded in:

```
# file: scripts/python/examples/testpybel.py

LINE 197: def testRFformaterror(self):
LINE 198:     """Test that invalid formats raise an error"""
LINE 199:     self.assertRaises(ValueError, self.RFformaterror)
```

Therefore, the testing process is not passed.

ghutchis:
I was aware of the "priority" of the sequence issue, which is why I said you could also just change the test. The modified version looks fine. 

I'll double-check the build tests, but this looks good to merge.

PR #2529, opened 2022-08-23
Include <ctime> for clock()

heirecka:
The code started to fail to build with gcc 12.2.0 here without this
patch, but including what is used should be a sensible thing anyway.


thesamesam:
Was this fixed by https://github.com/openbabel/openbabel/commit/c0570bfeb2d7e0a6a6de1f257cf28e7f3cac8739?

heirecka:
Oh, I didn't see that. But yes.

PR #2537, opened 2022-09-16
Fix 338

### LINKED ISSUE #338: Problem with the negative charges when reading .cdx
When reading a .cdx with negative charges, the charges get an unsigned number.  
For example, converting a given .cdx containing two charged atoms into the mol format, atom no.8's charge is set to 255 instead of being -1. 
Hence, I see:  
M  CHG  2   4   1   8 255
instead of : 
M  CHG  2   4   1   8  -1

I have modified the file chemdrawcdx.cpp as below to fix this when the condition cdxr.GetLen()==1 is True. However, I was not able to test the code found in the else statement since I never meet this case. 

```
case kCDXProp_Atom_Charge:
          if(cdxr.GetLen()==1) 
            charge = (int)((char)cdxr.data().get());
          else
            READ_INT32(cdxr.data(), charge);
```

fredrikw:
Fixes #338, basically following the proposal from the initial report. However, reading the specification at http://www.cambridgesoft.com/services/documentation/sdk/chemdraw/cdx/properties/Atom_Charge.htm the charge is an int8 so I cast directly to that.
Also according to the mentioned spec, there was a bug in ChemDraw producing a different kind of charges for a while. It looks like the code will not work on those files, the "else-path" should be invoked for those but according to the spec it looks like there should be a shift of the result to get a proper charge. But as long as we cannot find an example file created in ChemDraw >= 8.0 < 8.0.6 we will not be able to test for this.


robertwerner67:
*Back then* when ChemDraw was small enough to fit on a CD-ROM sold (and distributed as a trial version working a couple of days) by CambridgeSoft ...  On a more serious side for the specific range of versions of ChemDraw, though, it may be worth to get in touch with
+ Bartosz Kosiorek as he maintains the entry for ChemDraw on [winehq.org](https://appdb.winehq.org/objectManager.php?sClass=application&iId=3060) including one line for 8.0.3. (equally known as ChemDraw/Office 2004)
+ Tamas E. Gunda for that his [review](https://www.gunda.hu/dprogs/) of sketchers started (footnote 11) back in 1998 (in Hungarian) and 2000 (in English).

Maybe they know someone with a working copy, the .iso archived on a server, or the CD around.

aminophen:
Thanks for the fix!

PR #2539, opened 2022-09-23
If requested, supply atom index as atom classes

ghutchis:
Signed-off-by: Geoff Hutchison <geoff.hutchison@gmail.com>


baoilleach:
Hi @ghutchis, I don't know all the details but I think this is a mistake. I often deal with molecules where only some atoms have atom classes. With this change they will be written out with a mixture of atom classes and atom indices. Couldn't this behaviour be requested with an option, e.g. add indices as atom classes on writing?

ghutchis:
Sure, good point. I'll rework it.

PR #2540, opened 2022-09-28
document silent deps: maeparser and CoordgenLibs

ostueker:
Packagers need to know of these silent dependencies like maeparser and CoordgenLibs so that they can be properly managed. Otherwise re-building the package is not reproducible.


ghutchis:
Practically speaking, I'm going to dump both of them for now. The contributor who proposed integrating `coordgen` never finished that work, so it's an unused dependency.

ghutchis (approved):

PR #2551, opened 2022-11-23
ci: add CIFuzz Github Action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where Open Babel already runs (https://github.com/google/oss-fuzz/tree/master/projects/openbabel). CIFuzz can help detect regressions and moves the fuzzing a bit further into the development pipeline, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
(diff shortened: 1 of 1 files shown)


ghutchis (approved):
Sure - definitely worth a try

PR #2563, opened 2023-02-04
fix: vaspformat

e-kwsm:
- add WRITEONEONLY to Flags
- expand indentation tab
- fix alignment


src/formats/vaspformat.cpp
```diff
@@ -120,7 +120,7 @@ namespace OpenBabel {
        READBINARY  WRITEBINARY  READXML  ZEROATOMSOK */
     unsigned int Flags() override
     {
-      return READONEONLY;
+      return READONEONLY | WRITEONEONLY;
```
e-kwsm, line 123: I am not sure what is the *correct* value here, though,\.

PR #2565, opened 2023-02-06
correct an ommision, complete output option

nbehrnd:
Without the `x` in `-xm`, the output option is not going to work.

### DIFF
--- a/src/formats/json/chemicaljsonformat.cpp
+++ b/src/formats/json/chemicaljsonformat.cpp
@@ -240,6 +240,51 @@ namespace OpenBabel
       StereoFrom3D(pmol);
     }
 
+    // add the unit cell if present
+    if (inRoot.HasMember("unitCell") && inRoot["unitCell"].IsObject())
+    {
+      const rapidjson::Value &unitCell = inRoot["unitCell"];
+      if (unitCell.HasMember("a") && unitCell["a"].IsDouble())
+      {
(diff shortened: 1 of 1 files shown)


nbehrnd:
Because I have no working knowledge of any C language, the following only can be a suggestion for an additional feature here:

It took me some time to identify `babel_povray3.inc` in my file system (Linux Debian 12/bookworm), which I actually have twice; namely in path `/usr/share/openbabel/3.1.1` as well as in `/usr/share/povray/include`.  Maybe because I do use `openbabel` as provided by the Linux Debian 12/bookworm repository as well as `Avogadro2`, though I'm not sure if either one of the programs just wants to be sure that PovRay has a reference instructions for colours and dimensions to use.

But here is a potential pitfall: moving to an other computer, and forgetting the very `babel_povray3.inc` file.  Can we have an option that this file `.inc` file is copy-pasted into the current working directory when OpenBabel is requested to generate the first `.pov`  in the current working directory?  (This would be similar to e.g., Jmol; if one requests to export a structure as `.obj`, a complementary `.mtl` file is written on the fly, too.)  At least as an *optional* toggle?

fredrikw:
The errors are unrelated to the PR. The SWIG bindings failure is probably temporary, failed downloads of packages. The lint_python warns for an outdated version of setuptools with a vulnerability in it. I don't even think that the setuptools version is specified in OpenBabel so that is probably related to the GitHub action.

Issue #2374, opened 2021-05-28
Local installation of openbabel has a non-working OpenBabel3.cmake file


correaa:
- [x] I believe this to be a bug with Open Babel
- [ ] This is a feature request

## Environment Information
Open Babel version: 3.3.1
Operating system and version: Ubuntu 20.04

## Expected Behavior

Local installation of OpenBabel should have a working `~/lib/cmake/OpenBabel3.cmake`.

## Actual Behavior

Package configuration is found but then rejected later because of `maeparser` and `coordgen` not found.

```
  Found package configuration file:

    /home/correaa/lib/cmake/openbabel3/OpenBabel3Config.cmake

  but it set OpenBabel3_FOUND to FALSE so package "OpenBabel3" is considered
  to be NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: maeparser
  coordgen
```

I tried to install `maeparser` (git, cmake, make install) but I get the same error.

## Steps to Reproduce

```
git clone git@github.com:openbabel/openbabel.git
cd openbabel
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX:PATH=$HOME ..
make -j 10
make test
make install
```
In a different directory have a `CMakeLists.txt` with the contents

```
find_package(OpenBabel3 REQUIRED)
```
and run `cmake .`, gives:

```
CMake Error at CMakeLists.txt:30 (find_package):
  Found package configuration file:

    /home/correaa/lib/cmake/openbabel3/OpenBabel3Config.cmake

  but it set OpenBabel3_FOUND to FALSE so package "OpenBabel3" is considered
  to be NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: maeparser
  coordgen
```

PR #2574, opened 2023-02-23
OpenBabel3Config.cmake to keep include information

correaa:
this may solve https://github.com/openbabel/openbabel/issues/2374


ghutchis:
LGTM

PR #2580, opened 2023-03-29
use simple cube format for molecules with simple grid

correaa:


correaa:
I propose to default to the (simpler) format (a) in the case that there is only one grid associated with the molecule.
Having one grid is very common and it is also common that other programs only format (a) anyway.

Please let me know if for some reason this is not acceptable and what I can do to fix it.

PR #2583, opened 2023-04-08
AddAtom argument constness

correaa:
The function AddAtom can take argument by const.
The only two things that the function does with the argument are to GetIdx (which is const) and make a copy (clone) to insert in the molecule data structure.


romanbsd:
👍

PR #2596, opened 2023-06-12
Fix bug in pubchem json coordinate writing

magnusrahm:
In current master, when converting to pubchem json, x coordinates are always identical to y coordinates. This PR fixes that bug.


ghutchis:
Thanks. That's a pretty clear fix.

PR #2598, opened 2023-06-16
fix bug when read gro file as first residue idx is 0

DStarEpoch:
* when the first resid is 0, it will trigger object res without initialization
	* fix bug by checking whether object res is nullptr


ghutchis:
Sorry for the delayed merge - looks great, and thanks for the fix!

PR #2634, opened 2023-10-20
Fix bug in distance geometry code

timvdm:
This bug was discovered while testing out the new iterators. The old iterator marcos used iterators that had an operator bool() which implicitly converts into unsigned long. Instead of using the atom's id, a boolean value was used for IsCis/IsTrans.


ghutchis (approved):
Looks good to me.

PR #2637, opened 2023-10-24
Support for segment names in pdb files

dkoes:
This adds support for handling segment names in pdb files (they were previously ignored).  Importantly, atoms with different segments are now placed in different residues.  When building biological assemblies, Prody distinguishes between the mirrored chains of an assembly by setting their segment names.  Without this fix, you will end up with multiple residues within a single OBResidue.

### DIFF
--- a/include/openbabel/residue.h
+++ b/include/openbabel/residue.h
@@ -79,6 +79,8 @@ namespace OpenBabel {
     //! MODRES records for modified residues:
     //! http://www.rcsb.org/pdb/file_formats/pdb/pdbguide2.2/part_36.html
     void    SetName(const std::string &resname);
+    //! \brief Set the segment name of this residue (max four characters)
+    void    SetSegName(const std::string &segname);    
     //! Set the residue number (in the sequence)
     void    SetNum(const unsigned int resnum);
     void    SetNum(const std::string  resnum);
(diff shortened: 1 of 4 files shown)


ghutchis:
@dkoes - I'm seeing test failures in the new test:
> Traceback (most recent call last):
  File "/Users/runner/work/openbabel/openbabel/test/testpdbformat.py", line 28, in testSegname
    mol = pybel.readstring('pdb','''ATOM    102  N   CYS A  16      59.916  27.715  54.719  1.00 30.93         A N
AttributeError: 'NoneType' object has no attribute 'readstring'

dkoes:
Huh, why might that be?  I'm importing pybel with `from testbindings import pybel` which is what other test files do.  The test passes on my workstation and on CI.

fredrikw:
A quick look on the failing test seems to indicate that the new test is run also on installs without the bindings. (https://github.com/openbabel/openbabel/actions/runs/6626985692/job/18417945235?pr=2637)
So I would assume it is fixed by either rewriting the test to not use the bindings, similar to the other test in testpdbformat.py or split it out to a new file and only run the test when the bindings are built.
With option one being preferred I would think, since this should probably be a feature that should be available (and tested) also without bindings.

dkoes:
Thanks for explaining that to me.  I've added an obabel test and guarded the pybel dependent test with a check for the presence of pybel (need this test to actually test the API).  The new test revealed a bug with writing the segnames, so that has also been fixed.

PR #2648, opened 2023-11-23
Remove redundant SSSR rings based on bond paths

### LINKED ISSUE #2641: Unexpected behaviour in SSSR routine - wrong sub rings
Open Babel version: current
Operating system and version:  linux or macos (tested both) 

Looking at the following smiles: O1C23C4=C(C=C12)C1=COC(=C4OC3)C1

I would expect to see a SSSR containing 5 subcycles, which is what is seen, however the subcycles generated for each OBRing* shows some weird behaviour. 

running the following code on the Mol object : 

```c
unsigned int rc = 0;
FOR_RINGS_OF_MOL(r,mol){
  obring = &(*r);
  fprintf(stderr,"%d [ ",rc++);
  for(unsigned int i=0;i<obring->Size();i++){
    ratom = mol->GetAtom(obring->_path[i]);
    fprintf(stderr,"%d ",ratom->GetIdx());
  }
  fprintf(stderr,"]\n");
}
```

gives the following output: 

```
0 [ 6 1 2 ]
1 [ 3 4 5 6 2 ]
2 [ 12 11 3 2 13 ]
3 [ 7 8 9 10 14 ]
4 [ 3 4 5 6 1 2 ]
```

Looking at ring (3), we see atom indexes that are not shared with any other subcycle in the list? which would be impossible if the subcycles are generated as i would expect? To confirm, the method in RDKit does not show this behaviour: 

```python 
mol = rdkit.Chem.MolFromSmiles('O1C23C4=C(C=C12)C1=COC(=C4OC3)C1')
for sequence in rdmolops.GetSymmSSSR(mol):
    for i in sequence:
        print(f"{i} ", end='')
    print()
```
yields

```
[ 0 1 5 ]
[ 4 3 2 1 5 ]
[ 7 6 13 9 8 ]
[ 11 10 2 1 12 ]
[ 13 6 3 2 10 9 ]
```

Where overlap is seen for all the atoms across the ring set. 

Happy to attempt a fix on this but some direction would be appreciated (if in fact this is a true bug, and not my misunderstanding). I believe its due to the 3 cycle, as for the majority of cases i see the correct overlaps.

fredrikw:
As found in https://github.com/openbabel/openbabel/issues/2641#issuecomment-1816436930, the current SSSR has a problem when removing redundant rings. The overlap is currently based on atoms instead of bonds, leading to incorrect results.
This code includes a fix for that problem, but I do not know how efficient the solution is.

In addition, I have a few questions/problems regarding the testing. I added a test case to `ringtest.cpp` where I thought it belonged. However, since several other tests uses the file `test/files/attype.00.smi` this leads to a bunch of failing tests. Would it be best to rework "my test" to a regression test or regenerate the result files for the other failing tests (such as formalcharge and formula etc)?
In addition, this change seems to have affected canonicalization, at least I get a ton of errors in test_canonstable. If that code relies on SSSR it would be reasonable to expect this and in that case maybe a "backwards-compability-switch" so that the canonicalization could continue to call the old code that although not correct for SSSR-purposes should be good enough to produce stable canonicalization.


ghutchis:
Since this would be relevant for 3.2, and we can't guarantee canonicalization across feature changes, I think we're okay wrt backwards compatibility of canonicalization.

I would regenerate the result files as far as the tests.

fredrikw:
Is there an easy way to regenerate all result files?

fredrikw:
Weirdly enough, the smartstest is ok when I run it locally. I cannot find a good reason for that. I also get more changes than expected in the formalcharge test results file so something is up with that as well... I'll need to look more at this.

fredrikw:
Two fails in the tests. One is in testobconv_writers.py and I cannot see how that test has ever worked since the x- and y-coordinates in the expected output are the same. (https://github.com/openbabel/openbabel/blob/32cf131444c1555c749b356dab44fb9fe275271f/test/testobconv_writers.py#L2882-L2888)
The other is a bigger problem. It's in canonstabletest and the problem is that the failing test  there isn't stable.
```sh
$ BABEL_LIBDIR=build/lib/ build/bin/obabel -ismi -ocan -:"C1[C@@H]2C[C@H]3C[C@H]1C[C@](C2)(C3)Nc1nc(nc(n1)N[C@]12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3)NC12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3"
C1[C@H]2C[C@@H]3C[C@H]1C[C@@](C2)(C3)Nc1nc(nc(n1)N[C@]12C[C@@H]3C[C@@H](C2)C[C@@H](C1)C3)NC12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3
1 molecule converted

$ BABEL_LIBDIR=build/lib/ build/bin/obabel -ismi -ocan -:"C1[C@H]2C[C@@H]3C[C@H]1C[C@@](C2)(C3)Nc1nc(nc(n1)N[C@]12C[C@@H]3C[C@@H](C2)C[C@@H](C1)C3)NC12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3"
C1[C@@H]2C[C@H]3C[C@H]1C[C@](C2)(C3)Nc1nc(nc(n1)N[C@]12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3)NC12C[C@@H]3C[C@@H](C2)C[C@H](C1)C3
```

So somehow, there seem to be an edge case in the canonicalization algo that the new SSSR ring code breaks. I don't know the canon code well enough to understand why atm but I will try to look into it. But maybe someone more versed into it can help out?

fredrikw:
OK, found the source of the failing pcjson-test. There was a bug in that format, fixed in #2596. However, this test was not updated to match. I will add a PR for that...

fredrikw:
I found the problem to the failing canonical test. There was a bug in my new code when there were multiple ring systems in the molecule. It is now fixed and thereby I could also revert the changes to the canonstable file meaning that there are no longer any changes to the canonicalization.
So I think I am done with this now!

ghutchis:
@fredrikw - this looks good, esp. if it doesn't change canonicalization. I don't see the specific test case from the issue added anywhere, e.g. some SSSR test for O1C23C4=C(C=C12)C1=COC(=C4OC3)C1 to ensure we don't regress.

ghutchis:
Otherwise, the logic looks good to me, I'm ready to merge.

fredrikw:
There are tests added in test/files/attype.00.smi and test/files/ringresults.txt that should test this issue if I'm not mistaken.

PR #2661, opened 2023-12-24
Upgrade hinformat.cpp InterX

ileontyev:
The new HIN features include parsing residue information
- atomIDs (labels)
- chain
- comments (to OBGenericData, OBDaataPair)
- partial/formal charges,
and etc.

### DIFF
--- a/src/formats/hinformat.cpp
+++ b/src/formats/hinformat.cpp
@@ -23,7 +23,28 @@ GNU General Public License for more details.
 
 #include <cstdlib>
 
+#include <iostream>
+#include <openbabel/generic.h>
+
 using namespace std;
+  char* strcat_mem(const char *str1, const char *str2) {
+   char* result;
+    int l2 = strlen(str2) ;
+    if (!str1) {
+     result = (char*) malloc(l2+1);
+     memcpy(result, str2, l2) ;
+     result[l2] = '\0' ;
+     return result;
+    }
+    int l1 = strlen(str1) ;
+    result = (char*) malloc(l1 + l2 + 1);
+    if(!result) return result;
+    memcpy(result, str1, l1) ;
+    memcpy(result + l1, str2, l2 );
+    result[l1 + l2] = '\0' ;
+    return result;
+  }
+
 namespace OpenBabel
 {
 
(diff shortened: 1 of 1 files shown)


src/formats/hinformat.cpp
```diff
@@ -23,7 +23,28 @@ GNU General Public License for more details.
 
 #include <cstdlib>
 
+#include <iostream>
+#include <openbabel/generic.h>
+
 using namespace std;
+  char* strcat_mem(const char *str1, const char *str2) {
```
ghutchis, line 30: Why? Can't we use C++ string catenation? This seems like a bad idea.

src/formats/hinformat.cpp
```diff
@@ -34,16 +55,28 @@ namespace OpenBabel
...
-
     const char* Description() override  // required
     {
       return "HyperChem HIN format\n"
-             "No comments yet\n";
+        "Read Options e.g. -aR -aT\n"
+        "  R               Persive residue information (atom label, resname, chainID) automatically, ignoring info parsed from HIN. Works for PROTEIN only.\n"
+        "  T               Read atom Types from HIN (glitchy, because convertion formats usually Translate() types using predifined ttab).\n"
+        "                  By default, atom types from HIN are ignored, then auto persived at the converted format Write routine (e.g. to mol2 types)\n"
+        "  C               Parse pair-formatted comments (fmt: ;field-name value) to OBPairData type, otherwise multi-line Generic OBCommentData\n\n";
```
ghutchis, line 72: Parsing pair-formatted comments / data should be the default.

src/formats/hinformat.cpp
```diff
@@ -34,16 +55,28 @@ namespace OpenBabel
...
+      OBConversion::RegisterOptionParam("l", this, 0, OBConversion::OUTOPTIONS);
+      OBConversion::RegisterOptionParam("u", this, 0, OBConversion::OUTOPTIONS);
     }
-
     const char* Description() override  // required
     {
       return "HyperChem HIN format\n"
-             "No comments yet\n";
+        "Read Options e.g. -aR -aT\n"
+        "  R               Persive residue information (atom label, resname, chainID) automatically, ignoring info parsed from HIN. Works for PROTEIN only.\n"
```
ghutchis, line 69: "Perceive" - but this should default to true.

ghutchis (changes requested):
The concept is great, but please use C++ strings to catenate instead of a version of strcat.

Thanks.

PR #2664, opened 2024-01-16
Implemented code for atom-typing to support the Alexandria force field.

dspoel:
The Alexandria Chemistry Toolkit for building force fields will be released shortly. It relies on the OpenBabel software to support it for generating atom types. In this patch the forcefieldalexandria is introduced as a child class to forcefieldgaff in an effort to change as little code as possible.

There are many changes to bondtyp.txt which admittably will be hard to check. These will affect any force field typing in OpenBabel.


dspoel:
Hm, the test failure seems to be related to code I haven't touched. Any suggestion on how to debug this?

PR #2678, opened 2024-03-05
Make CorrectForPH respect perceived chains.

### LINKED ISSUE #2677: Adding hydrogens with -p messes up the entire structure (not only residue numbers and names)
First of all thanks for working on and maintaining OpenBabel!

I want to add hydrogens to my pdb file. If I use `obabel 3lcs.pdb -O 3lcs_prot.pdb -h` everything looks great in general but the results don't match my expectations - for example one of the oxygens in the side chain of glutamic acid (e.g. GLU 1197) gets a hydrogen. In physiological conditions (from my understanding) it doesn't have a hydrogen there and is charged.

So I moved on to use `-p` instead of `-h` to get hydrogens at pH 7.4. Now from what I can see the hydrogens are added following my expectation, but the pdb file is messed up.

(I am using OpenBabel within a Python project, but as the behavior is the same from the command line I used the command line options for easier reproducibility.)

Am I missing a preprocessing step or misunderstanding how these options should be used?

Thanks for your time and help!

## Environment Information
Open Babel version: 3.1.1
Operating system and version: Windows 10

## Expected Behavior
running `obabel 3lcs.pdb -O 3lcs_prot.pdb -p` should give me my pdb file with hydrogens added but residue numbers and names unchanged.

## Actual Behavior
running `obabel 3lcs.pdb -O 3lcs_prot.pdb -p` changes residue numbers to start at 1 and changes the residue names for everything it doesn't recognize as a standard amino acid or has some other issues with to UNK or UNL.
(From what I see alternative locations and missing atoms seem to cause issues and any ligands in the file also lose their name.)

## Steps to Reproduce
download [the pdb file](https://files.rcsb.org/download/3LCS.pdb) and run the commands (also happens with other pdb files though)

fredrikw:
This fixes #2677 
CorrectForPH currently clears all perceived attributes through mol.EndModify(). In my change, I check if chains are perceived before hand and in that case sets them as perceived also after since normally the residues and chains shouldn't change through pH-correction.
However, this is a breaking change if someone previously relied on the old behavior. In that case they need to manually reperceive the chains after the call to CorrectForPH (or reset the flag before...).


ghutchis (approved):
LGTM - makes total sense

ghutchis:
I'm finding a bunch of chain perception / EndModify bugs.

Issue #1996, opened 2019-06-23
Compilation failure


Vcald:
Does someone have a clue for such compilation failure on fedora 30? 

# cmake .. -DRUN_SWIG=on -DPYTHON_BINDINGS=on -DBUILD_GUI=on 
-- The C compiler identification is GNU 9.1.1
-- The CXX compiler identification is GNU 9.1.1
-- Check for working C compiler: /usr/lib64/ccache/cc
-- Check for working C compiler: /usr/lib64/ccache/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++
-- Check for working CXX compiler: /usr/lib64/ccache/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at CMakeLists.txt:12 (cmake_policy):
  The OLD behavior for policy CMP0042 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.

-- Using included inchi library.
-- Found LibXml2: /usr/lib64/libxml2.so (found version "2.9.9") 
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.3") 
-- Found wxWidgets: -pthread;;;-lwx_baseu-3.0;-lwx_gtk3u_core-3.0;-lwx_gtk3u_adv-3.0 (found version "3.0.4") 
-- Looking for conio.h
-- Looking for conio.h - not found
-- Looking for sys/time.h
-- Looking for sys/time.h - found
-- Looking for time.h
-- Looking for time.h - found
-- Looking for strings.h
-- Looking for strings.h - found
-- Looking for rpc/xdr.h
-- Looking for rpc/xdr.h - found
-- Looking for regex.h
-- Looking for regex.h - found
-- Looking for C++ include sstream
-- Looking for C++ include sstream - found
-- Looking for rint
-- Looking for rint - not found
-- Looking for snprintf
-- Looking for snprintf - found
-- Looking for sranddev
-- Looking for sranddev - not found
-- Looking for strcasecmp
[...]

PR #2684, opened 2024-03-22
Change check for rpc/xdr.h

fredrikw:
Change from checking if the file exists to if the symbol xdrstdio_create exists. Should fix https://github.com/openbabel/openbabel/issues/1996


fredrikw:
Maybe we should also remove the extra check for CYGWIN on line 235. If I understood that comment correctly, this change should fix that bug as well, in a less definitive way (in case CYGWIN adds the missing symbol in the future...). But I don't have access to CYGWIN to test.

ghutchis:
I think for now, I'm going to merge as-is and see if I can get a GitHub build for Cygwin. (Not sure how important that platform is with WSL now.)

PR #2693, opened 2024-05-07
Fix memory leaks for xmlChar allocated by libxml

romanbsd:


romanbsd:
@ghutchis can you please take a look?

ghutchis (approved):
LGTM, thanks

PR #2694, opened 2024-05-07
Support conversion to CMLR

romanbsd:
Some formats read a reaction to OBMol using OBReactionFacade. Convert OBMol to OBReaction so that CML Reaction format can handle it.


ghutchis (approved):
LGTM, thanks

PR #2702, opened 2024-06-21
Restore compatibility with libxml 2.12.7

merkys:
This PR fixes the single incompatibility with libxml 2.12.7 which [has been reported on Debian package](https://bugs.debian.org/1073327). Please note this PR does not update the libxml sources as included under `include/libxml/`.

### DIFF
--- a/src/formats/xml/xml.cpp
+++ b/src/formats/xml/xml.cpp
@@ -242,13 +242,12 @@ namespace OpenBabel
 
     if(result==-1)
     {
-      xmlError* perr = xmlGetLastError();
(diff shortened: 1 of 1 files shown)


ghutchis (approved):

ghutchis:
The Python lint error is a separate issue, so lgtm.

merkys:
Thanks for merging this @ghutchis. I do not know much about libxml, but it seems that `xmlResetError()` is no longer necessary as the main thread is responsible for keeping and cleaning up the last XML error.

PR #2721, opened 2024-09-16
Fetch more recent rapidjson codebase rather than last release

matterhorn103:
Version 1.1.0 of rapidjson is now 8 years old, and building Open Babel is currently broken on recent Linux systems as a result. Using the current code incorporates the many patches since 2016 such as [this PR](https://github.com/Tencent/rapidjson/pull/719).


ghutchis:
No, we're not going to grab `master`. For consistency and security, it's much better to grab a specific commit hash. That's immutable.

I'd rather see a new release of rapidjson, but in the meantime, I'd prefer something like `7c73dd7de7c4f14379b781418c6e947ad464c818`

matterhorn103:
Fair enough, I swapped in the hash you suggested i.e. the most recent commit hash.

PR #2737, opened 2024-12-07
Implement fuzz_convert fuzzer

tyler92:
A new fuzzer is implemented - fuzz_convert. It tests `OBConversion::Convert` function will randomized input/output formats and data. It will help to find many crashes - some of them I will report one by one later, but we already can merge this PR to increase coverage. Additional changes in this PR:

1. src/data.cpp: oot of bound read in `OBGlobalDataBase::Init()`
2. src/formats/MCDLformat.cpp: usage of uninitialized variable `numdups` that leads to OOM or crashes
3. src/formats/adfformat.cpp: fix infinite loop in case of unexpected end of input
4. src/formats/abinitformat.cpp: prevent division by zero
5. src/formats/gulpformat.cpp: fix infinite loop in case of unexpected end of input
6. src/formats/xsfformat.cpp: prevent division by zero

### DIFF
--- a/src/data.cpp
+++ b/src/data.cpp
@@ -704,7 +704,7 @@ namespace OpenBabel
                 strncpy(charBuffer, p1, (p2 - p1));
                 charBuffer[(p2 - p1)] = '\0';
                 ParseLine(charBuffer);
-                p1 = ++p2;
+                p1 = p2 + 1;
               }
         }
       else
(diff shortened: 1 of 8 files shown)


ghutchis:
I have no doubt that many formats have bugs. Thanks, LGTM.

PR #2753, opened 2024-12-12
build!: use CMake FindPython

e-kwsm:
FindPythonInterp and FindPythonLibs are deprecated since CMake 3.12; use FindPython instead

- https://cmake.org/cmake/help/v3.12/module/FindPython.html
- https://cmake.org/cmake/help/v3.12/module/FindPythonInterp.html
- https://cmake.org/cmake/help/v3.12/module/FindPythonLibs.html

This requires CMake 3.12 or later


ghutchis (approved):
Thanks, this is greatly appreciated.

ghutchis (approved):
LGTM, thanks

Issue #2763, opened 2025-01-09
pybel readstring produces incorrect molecule for 'cdjson' format


gregmagdits:
## Environment Information
Open Babel version: 3.1.0
Operating system and version: Ubuntu (WSL)

## Expected Behavior
readstring('cdjson','''{"m":[{"a":[{"x":0,"y":0,"i":"a3"}],"i":"m3"}]}''').write('smi').strip()  should return 'C'

## Actual Behavior
The above code snippet produces '[C]'

## Steps to Reproduce

The following code snippet in python demonstrates the issue:
```python
# this passes
assert readstring('smi', 'C').write('smi', opt={}).strip() == 'C'
mol = readstring('smi', 'C')
# add 2D coordinates necessary for cdjson
mol.make2D()
cd_json = mol.write('cdjson', opt={})

# bug is here since the cd_json above seems to be a correct representation of methane according to the chemdoodle json spec - https://web.chemdoodle.com/docs/chemdoodle-json-format 
converted_smiles =readstring('cdjson', cd_json, opt={}).write('smi', opt={}).strip()

# This assert fails because converted_smiles == '[C]' instead of 'C'
assert  converted_smiles == 'C'
```

It appears that there is a bug in the readstring procedure when deserializing chem-doodle json, since the molecule returned in the above code snippet has a formula of C instead of CH4

fredrikw:
I was following up on this and it seems that there are plenty of problems with the cd_json format. I tried a few round-trips that failed pretty bad:
```
➜ obabel -ismi -:C -ocdjson --gen2D | obabel -icdjson -osmi
[C]
➜ obabel -ismi -:CC -ocdjson --gen2D | obabel -icdjson -osmi
[C][C]
➜ obabel -ismi -:"C(=O)CO" -ocdjson --gen2D | obabel -icdjson -osmi
[C](=[C])[C][C]
```
The last one is both a problem with generating and reading cdjson since the cdjson from the first step seems to produce atoms with the label "8" instead of oxygens.

PR #2769, opened 2025-01-21
Fix cdjson

gregmagdits:
updated chemdoodlejsonformat.cpp ('cdjson' format) to fix issues described in #2763


ghutchis:
LGTM, thanks - and apologies for the long delay.

PR #2773, opened 2025-01-29
Test fixes

fredrikw:
Fixes the failing tests in pybindtest_obconv_writers and pybindtest_bindings. The former by relaxing the number of decimals tested in json outputs and the second by not requiring Cairo (_png2-format) for pybel to run `mol.draw(show=False, update=True)`.

I also looked at the failing test on Mac (test_cifspacegroup_11) but that problem seems to be more involved, the output is very different from the expected one and I don't know spacegroups enough to understand what is happening.

### DIFF
--- a/scripts/python/openbabel/pybel.py
+++ b/scripts/python/openbabel/pybel.py
@@ -708,7 +708,7 @@ class Molecule(object):
 
             filedes, filename = tempfile.mkstemp()
 
-        if show or filename:
+        if show or (filedes is None):
             workingmol.write("_png2", filename=filename, overwrite=True)
 
         if show:
--- a/scripts/python/openbabel/pybel.py
+++ b/scripts/python/openbabel/pybel.py
@@ -672,12 +672,13 @@ class Molecule(object):
         Tkinter and Python Imaging Library are required for image display.
         """
         obconversion = ob.OBConversion()
-        formatok = obconversion.SetOutFormat("_png2")
-        if not formatok:
-            raise ImportError("PNG depiction support not found. You should "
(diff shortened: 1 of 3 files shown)


ghutchis:
Thanks - unfortunately, I've been swamped with work, but I'll take a look at the Mac test soon.

fredrikw:
Maybe we should also look at the tmate functionality. I don't think it is used often and it delays the test reports a lot. I don't know if it is possible to only run it on demand, e.g. by setting some flag or something...

ghutchis:
Yes, you add something like this to the GitHub action:

```
    - name: Setup tmate session
      if: ${{ failure() }}
      uses: mxschmitt/action-tmate@v3
```

ghutchis:
The reason `tmate` delays the test results is that on failure, you have ~30 minutes(?) to `ssh` to the runner using `tmate` and investigate. (It might be longer, I don't remember.)

fredrikw:
I just looked at the documentation for the tmate-action and it looks like there is a way to trigger it manually: https://github.com/mxschmitt/action-tmate?tab=readme-ov-file#manually-triggered-debug
What do you think about that? Or we could shorten the timeout, but I would prefer the manual trigger.

fredrikw:
I looked again at the Mac fail and it seems to be unrelated to the actual test... I think the problem is that the residue numbering is different for some reason, the occupation data is correct.

fredrikw:
The json normalization has been separately implemented in #2876 so I reverted that part.

ghutchis:
Thanks - I was looking through this trying to see what else was needed.

ghutchis (approved):
LGTM, thanks

Issue #2700, opened 2024-05-27
Formal releases or just continue to use the master branch


cwolcott:
## Environment Information
Open Babel version: 3.1.1
Operating system and version: Oracle Enterprise Linux 8.9

## Expected Behavior
3.1.1 was released in May 2020, and the contributors are doing great work continuing to update the master.  Will there ever be formal releases again.

jakobandersen:
It would really be nice to have formal releases. The 3.1.1 release is not compatible with the newest C++ versions, so compiling against it is not possible when using the default compiler on, e.g., macOS.

Homebrew by default seems to compile the newest release (https://formulae.brew.sh/formula/open-babel), so if one just installs that, downstream libraries fail to compile. They luckly also has the HEAD version, which works.
However, for conda-forge (https://anaconda.org/conda-forge/openbabel) I don't see a similar HEAD package, so unless I'm missing something, then it's not possible to create a Conda package that depends on Open Babel.
Other packagers probably have similar processes where it is formal releases they go by.

Even with various tricks, having a formal release seems to make all this much easier.

ghutchis:
There will be a new release of 3.2 by the end of May. Please follow #2889 for progress.

PR #2784, opened 2025-03-26
cmake4 compat

PPN-SD:
* up cmake_minimum_required to 3.10
* delete CMP0042 because OLD behavior is removed in cmake-4 and is no longer used.
* set ENABLE_EXPORTS for test_runner (OLD behavior for CMP0065 is removed in cmake4)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
-# Please ensure that any changes remain compliant with 3.1.
+# Please ensure that any changes remain compliant with 3.10.
 if(NOT EMBED_OPENBABEL)
-  cmake_minimum_required(VERSION 3.1)
+  cmake_minimum_required(VERSION 3.10)
 endif()
 
 project(openbabel)
@@ -10,9 +10,6 @@ set (CMAKE_CXX_STANDARD 11)
 
 if(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
-  if(POLICY CMP0042)
-    cmake_policy(SET CMP0042 OLD)
-  endif()
 endif()
 
 include (CheckCXXCompilerFlag)
--- a/doc/examples/static_executable/CMakeLists.txt
+++ b/doc/examples/static_executable/CMakeLists.txt
@@ -25,7 +25,7 @@
 #
 
 # This line is required for cmake backwards compatibility.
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Please ensure that any changes remain compliant with 3.1.
```
berquist: This comment needs to be updated.

e-kwsm (commented):
The following should be updated as well:

- https://github.com/openbabel/openbabel/blob/889c350feb179b43aa43985799910149d4eaa2bc/doc/examples/static_executable/CMakeLists.txt#L28
- https://github.com/openbabel/openbabel/blob/889c350feb179b43aa43985799910149d4eaa2bc/scripts/CMakeLists.txt#L1

PPN-SD:
That was already done for scripts/CMakeLists.txt

Example is just an example. But ok.

jakobandersen:
This is quite important to get fixed. Due to the lack of formal releases (https://github.com/openbabel/openbabel/issues/2700) one is on macOS forced to install the HEAD version, i.e., from the master branch. But now that doesn't work due to the new CMake version. So I'm not sure there is an easy way to install Open Babel on macOS via Brew any more.

hsk17:
while the commit from this PR allows building openbabel using cmake-4, it has an important side effect resulting in failure of tests charge_mmff94 and charge_gasteiger:  raising cmake_minimum_required to a value of 3.4 or higher triggers CMP0065, and then test_runner is linked such that these tests fail immediately when started.

openbabel/test/CMakeLists.txt needs an additional instruction:
set_target_properties(test_runner PROPERTIES ENABLE_EXPORTS TRUE)

PPN-SD:
> while the commit from this PR allows building openbabel using cmake-4, it has an important side effect resulting in failure of tests charge_mmff94 and charge_gasteiger: raising cmake_minimum_required to a value of 3.4 or higher triggers CMP0065, and then test_runner is linked such that these tests fail immediately when started.
> 
> openbabel/test/CMakeLists.txt needs an additional instruction: set_target_properties(test_runner PROPERTIES ENABLE_EXPORTS TRUE)

I can reproduce failures and fix. Thanks.

ghutchis:
I'm going to merge this and fix the test failures on a new branch. Thanks

PR #2796, opened 2025-05-22
output data to pdbqt format

dkoes:
For PDBQT format, output molecular data if present (likely as REMARK fields).

This is to address gnina/gnina#311

The code is copy and pasted from `pdbformat.cpp`, which is bad, but there doesn't seem to be a logical place to put shared code.


ghutchis (approved):
LGTM, thanks

PR #2798, opened 2025-05-28
Update pdbqtformat.cpp to match Meeko PDBQT VDW radius

bmp192529:
The current tool recommended by the AutoDock team for ligand preparation is Meeko. Currently, OpenBabel is able to produce pdbqt files extremely similar to Meeko, with the key difference being a vdw radius of 0.00 instead of 1.00. Example of the same molecule prepared in obabel vs Meeko:
[pdbqt-format-comparison.zip](https://github.com/user-attachments/files/20493907/pdbqt-format-comparison.zip)


dspoel:
Should not the vdw radius be a meaningful number rather than 0 or 1?

ghutchis:
A quick search indicates that MDAnalysis thinks that's an occupancy field?
https://userguide.mdanalysis.org/stable/formats/reference/pdbqt.html

bmp192529:
Hello all,Thank you for your quick response. The example pdbqt included in the official AutoDock guide from Scripps identifies the value as 1 for all atoms at least in their example file on page 29: https://autodock.scripps.edu/wp-content/uploads/sites/56/2021/10/AutoDock4.2.6_UserGuide.pdf
Thanks again.On May 29, 2025, at 10:24 AM, Geoff Hutchison ***@***.***> wrote:﻿ghutchis left a comment (openbabel/openbabel#2798)
A quick search indicates that MDAnalysis thinks that's an occupancy field?
https://userguide.mdanalysis.org/stable/formats/reference/pdbqt.html

dspoel:
An occupancy of 1.0 makes sense, this is the definition of the field for a pdb file. It can be < 1 in case there is not enough electron density or for alternate conformations. So the patch makes sense.

bmp192529:
What else needs to happen before this pull request can be accepted?

ghutchis (approved):
LGTM, thanks

Issue #170, opened 2015-05-01
Register Option Param and PointCloud


baoilleach:
<pre>
>obabel -:O[C@@H](C)N -oinchi
==============================
*** Open Babel Error  in OpenBabel::OBConversion::RegisterOptionParam
  The number of parameters needed by option "r" in Point cloud on VDW surface di
ffers from an earlier registration.
==============================
*** Open Babel Error  in OpenBabel::OBConversion::RegisterOptionParam
  The number of parameters needed by option "x" in Point cloud on VDW surface di
ffers from an earlier registration.
==============================
*** Open Babel Error  in OpenBabel::OBConversion::RegisterOptionParam
  The number of parameters needed by option "p" in Chemical Markup Language diff
ers from an earlier registration.
InChI=1S/C2H7NO/c1-2(3)4/h2,4H,3H2,1H3/t2-/m0/s1
1 molecule converted
</pre>

Reinis:
I see the same issue with other formats too:
```
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "p" in VDW surface in binary STL format suitable for 3D printing differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "c" in VDW surface in binary STL format suitable for 3D printing differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "p" in Point cloud on VDW surface differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "x" in Point cloud on VDW surface differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "s" in ChemKin format differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "s" in Macromolecular Crystallographic Info differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "s" in FASTA format differs from an earlier registration.
==============================
*** Open Babel Error  in RegisterOptionParam
  The number of parameters needed by option "r" in SMILES format differs from an earlier registration.
```

mj-harvey:
Certainly in the pointcloud and stl classes I only register those options once, each with one parameter. 
Are options registered into a single global set, rather than residing within the scope of their format?
[...]

PR #2814, opened 2025-08-03
feat(stl): add support for bond visualization to STL export

TheRealSami:
# Description

This pull request enhances the STL export functionality in `openbabel` by visualizing the bonds between atoms.
Previously, the STL output represented only atoms as spheres. With this change:

- Bonds can be visualized as cylinders connecting the bonding atoms (`-xb` option),
- Bond radius and spacing can be adjusted (`--bond-radius`/`--bond-spacing` options), and
- Bonds can be coloured according to [CPK coloring](https://en.wikipedia.org/wiki/CPK_coloring) (`-xc` option).

TLDR; This PR adds support for the [Ball-and-stick model](https://en.wikipedia.org/wiki/Ball-and-stick_model) in the STL export. 

# Implementation notes

- Implementation is limited to `stlformat.cpp`
- Unfortunately, testing was only performed locally since no tests for the STL format were found :(
- Added options: `--bond-radius`, `--bond-spacing`
- Modified options:
  - `-xc` now colours atoms and bonds (if the bond visualization is enabled)
  - `-xs <scaling_factor>` scales atom and bond size (if the bond visualization is enabled)

# Examples

| Molecule | No bonds [1] | With bonds [2] | With bonds + colour [3] |
|:--------:|:------------:|:--------------:|:-----------------------:|
| FH       | fh | fh_bonds | fh_bonds+color |
| C2H4     | c2h4 | c2h4_bonds | c2h4_bonds+color |
| [Caffeine](https://pubchem.ncbi.nlm.nih.gov/compound/Caffeine) | caffeine |  caffeine_bonds | caffeine_bonds+color |

> Examples are screenshots of  [viewstl.com](https://www.viewstl.com/#!)'s render of the models (unfortunately, wihtout any anti-aliasing)

[1]: No additional options
[2]: Options `-xs 0.2 -xb`
[3]: Options `-xs 0.2 -xb -xc`

Please let me know if there are any changes or additions you'd recommend!

### DIFF
--- a/src/formats/stlformat.cpp
+++ b/src/formats/stlformat.cpp
@@ -68,16 +68,22 @@ namespace OpenBabel
   class STLFormat : public OpenBabel::OBMoleculeFormat
   {
     public:
-      // Non-standard STL extension for colour: 5 bits per col, bit 15 set
-      template<uint8_t Red, uint8_t Green, uint8_t Blue> 
-      static constexpr uint16_t colour() {
-        return 0x800 | ((Red & 0x1F) << 10) | ((Green & 0x1F) << 5) | (Blue & 0x1F);
-      };
+      /**
+        * @brief Enumeration of supported non-standard color channel orders for STL.
+        *        See https://en.wikipedia.org/wiki/STL_(file_format)#Binary.
+        */
+      enum class ColourOrder : uint8_t { RedGreenBlue, BlueGreenRed };
 
+    public:
       /**
        * @brief Non-standard STL value for no color for triangle face.
        */
       static constexpr uint16_t NoColor{0};
+
+      /**
+       * @brief Default colour order.
+       */
+      static constexpr ColourOrder DefaultColourOrder{ ColourOrder::BlueGreenRed };
 
       /**
        * @brief Default value for -xc option. (Unit: A)
@@ -110,6 +116,7 @@ namespace OpenBabel
         */
         OBConversion::RegisterOptionParam("bond-size", this, 1, OBConversion::GENOPTIONS);
         OBConversion::RegisterOptionParam("bond-radius", this, 1, OBConversion::GENOPTIONS);
+        OBConversion::RegisterOptionParam("colour-order-rgb", this, 0, OBConversion::GENOPTIONS);
       }
 
       /// Return description.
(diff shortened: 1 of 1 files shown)


ghutchis:
Wow, looks great! As someone who has gotten into 3D printing molecular models, this would greatly simplify the workflow. As is, I've used VRML, imported into Blender then saved as STL in Blender for printing.

ghutchis:
BTW for colors, you should use the element table:
```
        double r, g, b;
        OBElements::GetRGB(atom->GetAtomicNum(), &r, &g, &b);
```

TheRealSami:
Hey @ghutchis thanks for the feedback, I'm happy that this useful to someone else as well! :)
As for the colours: commit https://github.com/openbabel/openbabel/pull/2814/commits/f451be18cfbfdd363563d93bc83fc48a7371929e now  uses `OBElements::GetRGB` rather than assigning custom colours for the STL export.

Also i've found out that my test website [viewstl.com](https://www.viewstl.com/#!) swaps the red and blue colour channels which leads to weird colours. Apparently this is an inherent problem with the STL file format since the color encoding isn't standardized - see [wikipedia stl file format](https://en.wikipedia.org/wiki/STL_(file_format)).
Since there is no way to automatically detect the correct colour channel ordering i also added the `--colour-order-rgb` flag to enable support for both RGB and BGR colour channel ordering with BGR being the default.

| Caffeine with colour ordering RGB | Caffeine with colour ordering BGR |
|:--------------------------------------------------:|:--------------------------------------------------:|
| caffeine_color_ordering_rgb | caffeine_color_ordering_bgr |

> Note: According to `openbabel`'s default colour scheme Oxygen should be red-ish and Nitrogen should be blue-ish

ghutchis:
Thanks. We will likely switch to C++17 after the next release, but it'll be a bit before we move to C++20.

There seems to be an un-related test failure that I'll investigate.

fredrikw:
Looks great! Shouldn't the new options be Write options for consistency? And maybe shorter to match all the other format specific options.

TheRealSami:
Hey @fredrikw , thanks for the feedback!
Yes, you are right ideally the options would be defined as write options.
Unfortunately, `obabel` has some technical limitations making this tricky in practice:

1. (write) options are registered globally for all formats, which causes warnings at runtime if the same option is registered with differing argument counts, and
    - References: https://github.com/openbabel/openbabel/issues/170 , [obconversion.cpp#L1675](https://github.com/openbabel/openbabel/blob/master/src/obconversion.cpp#L1675)
2. write/ options are limited to single characters
    - Reference: [obabel.cpp#L266](https://github.com/openbabel/openbabel/blob/master/tools/obabel.cpp#L266)

Because of this, I went with longer, format-specific option names defined directly in `stlformat.cpp`, since it felt more user-friendly and maintainable.

That said, I'm open to adjusting the option names if you have suggestions that would better balance consistency with practicality. :)
Happy to update if there's a preferred convention here!

ghutchis (approved):
This looks good - sorry for the long delay - I haven't had much time I could devote to openbabel.

PR #2840, opened 2025-11-17
Update to new version numbering scheme for (Culgi) COF file format.

pbecherer:
The Culgi software (now called Simcenter Culgi, after acquisition by Siemens) has changed the format of its version number from major.minor.patch (10.0.0, 10.0.1, 11.0.0 etc.) to "YYMM" plus an extra number for patch releases: 2511, 2511.0001, 2511.0002, 2605 etc. The old way of reading the version number in cofformat.cpp is too strict to handle that other version numbering format.

**This change will write the version number in the new style, and reads both the old and the new-style version number without problems. Test files were modified accordingly.**

Note that currently the Simcenter Culgi software itself still writes the version number in .cof files in the old style, so that they can be read with the current released version of Open Babel. Our intention is to change that once an official Open Babel release is out that can read the new-style version numbers.

(I have permission from my employer to contribute this (very small) change under the GPL 2.0 license, please contact me if any further proof or information is required).


ghutchis (approved):
LGTM, thanks

PR #2847, opened 2026-01-07
Handle getline return values

pranavk:
Newer compilers emit warning with -Wunused-value. This also makes these functions fail early when things go wrong.


ghutchis (approved):
LGTM, thanks

PR #2874, opened 2026-04-29
Extend fuzzing with 3 new harnesses

DavidKorczynski:


ghutchis:
Thanks. I'm coming back since no other maintainers stepped forward. I'd rather take these in steps. So SMARTS and molecules sounds good. The 2D depiction code is a minefield and I'm not sure that it's actually used much. Most people I know use RDKit for depiction.

I imagine these will release a flood of fuzz hits, so my request is:
- split SMARTS out into a separate PR
- drop depict for now (and we need to decide whether we continue supporting it)
- I'll merge molecules

There are probably a bunch of other useful operations to add to that OBMol fuzzer, but it's a good start.

ghutchis:
Despite CIFuzz failing - that's kind of the point. I fixed the first bug in another PR and I'll see what OSSFuzz turns up tonight.

ghutchis (approved):

PR #2875, opened 2026-05-08
Minor modification for CMake 4 compatibility

ghutchis:


ghutchis:
Removing in favor of #2784

PR #2901, opened 2026-05-13
build(coordgen): pin to v3.0.2

e-kwsm:


ghutchis:
Looks good - I'll check back after dinner, but I'm guessing it'll build cleanly. Thanks!

ghutchis:
Mac is giving a CMake error:
> -- Downloaded Coordgen 'v3.0.2' to >/Users/runner/work/openbabel/openbabel/external/coordgen-v3.0.2.
>CMake Error at external/coordgen-v3.0.2/coordgen/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

>  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

>  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

e-kwsm:
Thanks.
Unfortunately, the fix is not released yet, so I pinned it to the latest commit.

ghutchis:
The CIFuzz is a separate bug - I'll handle that in a different PR. This looks fine to merge.

PR #2920, opened 2026-05-15
fix: remove register storage class specifier

e-kwsm:
register is deprecated in C++11 and removed in C++17


ghutchis (approved):
Yes, definitely. Thanks!

PR #2926, opened 2026-05-17
Fix dangling reference in CML reaction writer

romanbsd:
Store the stringstream contents by value before searching the generated CML, rather than binding a const reference to the temporary returned by str().


ghutchis (approved):

ghutchis:
Not sure about the lint_python fail, but that's obviously unrelated

PR #2932, opened 2026-05-19
refactor testbabel

e-kwsm:
- refactor(testbabel.run_exec): specify universal_newlines=True in subprocess.Popen
- refactor(testbabel.run_exec)!: replace subprocess.Poepn with subprocess.run (python 3.5+)
- refactor(testbabel.run_exec): avoid arbitrary argument lists
- refactor(testbabel.run_exec): split argument beforehand
- refactor(testbabel.TestOBabel.testMissingPlugins): use env parameter; avoid modifying os.environ directly and shell=True

### DIFF
--- a/test/testRInChI.py
+++ b/test/testRInChI.py
@@ -53,18 +53,18 @@ class TestReactionInChIWriter(BaseTest):
         for eqm in [False, True]:
             for rsmi, rinchi in data:
                 if eqm:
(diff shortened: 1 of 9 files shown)


ghutchis:
It looks good - I guess my main concern is that we're not running tests on Windows and there were some win-specific fixes.

ghutchis (approved):
LGTM, thanks

PR #2935, opened 2026-05-19
Fix rsmi to rxn conversion

romanbsd:
I noticed that you're reviewing PRs quickly now, so I'm opening PRs with all the patches I accumulated. This was for fixing a real problem we had, but I think it's generic enough.

### DIFF
--- a/src/formats/rsmiformat.cpp
+++ b/src/formats/rsmiformat.cpp
@@ -16,12 +16,15 @@ GNU General Public License for more details.
 #include "openbabel/babelconfig.h"
 #include <string>
 #include <iomanip>
+#include <memory>
 #include <openbabel/mol.h>
 #include <openbabel/atom.h>
 #include <openbabel/elements.h>
+#include <openbabel/op.h>
 
 #include "openbabel/obconversion.h"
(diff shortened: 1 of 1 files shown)


ghutchis:
I'm not sure I understand the point of this as a specific case. For non-reactions you use --gen2d and --gen3d. Rather than create a specific rsmi flag, wouldn't the better thing be to tweak --gen2d and --gen3d to be aware of reaction conversions and act accordingly? 

That seems like it would be more consistent.

ghutchis:
I thought about this more and it seems like the wrong approach. Let's say someone just wants to read some reaction SMILES (filtering, etc.) - this patch now requires coordinates be created every read?

I'd much rather fix / adjust gen2d and gen3d to handle reactions if supplied, or have a reaction-specific op class.

romanbsd:
I'm opening some PRs with the patches I accumulated, as I'm leaving the chemistry project soon, and I thought that it might be of some use to others. Do what you think is right, I won't be offended :) opened another PR today with tautomers support in Ruby wrapper.

ghutchis:
Of course - pull requests are definitely welcome

PR #2938, opened 2026-05-19
Fix another round of OSS-Fuzz bugs

ghutchis:


ghutchis:
I'm going to merge this - the Ubuntu ARM fail is because that builder can't reach the network.

PR #2943, opened 2026-05-20
Tautomers support in Ruby wrapper

romanbsd:
Adds OBMol#tautomers which returns Enumerators and OBMol#canonical_tautomer which returns OBMol


ghutchis (approved):
Sure, looks fine

PR #2947, opened 2026-05-20
Add InChI library version (CURRENT_VER) to format description

ghutchis:


ghutchis:
@nbehrnd - this should solve your request - now gives the version in the InChI description:
`obabel -H inchi`

nbehrnd:
@ghutchis This will be great in perspective of now more frequent updates of openbabel, and the InChI project.  The later project equally advances; on the demo page beta versions either about enhanced stereochemistry, or improved Molecular Inorganics yield an `InChI=1B/...` instead of `InChI=1S/...` as in the current stable release 1.07.5.

PR #2951, opened 2026-05-23
fix(SpaceGroup)!: operator!= should return bool

e-kwsm:


ghutchis:
Thanks - good catch. I'll need to think about this a bit. I think it just changes the ABI, which will happen anyway with the 3.2 release.

PR #2952, opened 2026-05-23
refactor!: use std::erf and std::erfc (c++11)

e-kwsm:


ghutchis (changes requested):
I doubt anyone is using this header, but for a 3.2 release it looks bad to drop API.

I'd suggest we convert math/erf.h to simply use std, mark with OB_DEPRECATED and then of course the src/charges users are good.

ghutchis:
Otherwise, thanks - this is a good idea.

ghutchis (approved):
Thanks - I think this is a good approach even if no one actually uses these.

PR #2953, opened 2026-05-23
fix: add forgotten `throw`

e-kwsm:


ghutchis (changes requested):
This doesn't seem like a good idea. These are intended as minor optimization warnings.

This is causing all sorts of errors in the distance geometry tests since we hit some of these.

What's the benefit to patching this?

e-kwsm:
At the moment the expressions have no effects, so they can be removed, alternatively enclose them in `#ifndef NDEBUG`&ndash;`#endif`, or replace them with assert.

ghutchis:
Okay, but it's not even our code. If you want to use `#ifndef NDEBUG` that's okay. But I definitely don't think we want assertions or exceptions here - these are warnings that can come up with distance geometry, which is why the tests are failing.

Thanks

include/LBFGS/LineSearchBacktracking.h
```diff
@@ -49,15 +49,15 @@ class LineSearchBacktracking
 
         // Check the value of step
         if(step <= Scalar(0))
-            std::invalid_argument("'step' must be positive");
+            throw std::invalid_argument("'step' must be positive");
```
e-kwsm, line 52: step is checked afterword
https://github.com/openbabel/openbabel/blob/aa81cefd492e4fc8c47c2eedcb98c9351cc82eba/include/LBFGS/LineSearchBacktracking.h#L103-L104
so I believe early return is OK.

PR #2960, opened 2026-05-26
fix: add a missed header inclusion (-Wdelete-incomplete)

e-kwsm:
```
src/obconversion.cpp:703:21: error: deleting pointer to incomplete type
      'OBBase' is incompatible with C++2c and may cause undefined behavior [-Wdelete-incomplete]
```

https://github.com/openbabel/openbabel/blob/aa81cefd492e4fc8c47c2eedcb98c9351cc82eba/src/obconversion.cpp#L703


ghutchis (approved):
LGTM, thanks

PR #2964, opened 2026-05-27
CMake: Don't look for wxWidgets unless the GUI is requested

berquist:
wxWidgets is only used when building the GUI.  Don't even bother looking for it when building the GUI.  This was causing problems with system wxWidgets infecting an non-isolated environment that explicitly didn't have wxWidgets installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,15 +175,6 @@ if(ZLIB_FOUND)
   endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


ghutchis (approved):
LGTM, thanks

ghutchis:
I have a strong suspicion that there will promptly be a 3.2.1, etc. with a bunch of these fixes.

PR #2969, opened 2026-05-28
Fix headers with system InChi

PPN-SD:
I guess the bundled header would still have been used before 3e6cb82 if include/inchi hadn’t been cleaned up by package maintainer.


ghutchis (approved):
LGTM

PPN-SD:
Actually disable inchi (rm include/inchi && WITH_INCHI=NO) is not possible given inchidescriptor.cpp is always built. It was transparent until now because of the header path as explained above.
Is there a meaning in making descriptor.cpp optional?

ghutchis:
I'm not sure what you're asking. You want to turn off the inchi descriptor if WITH_INCHI=NO correct?

PPN-SD:
> You want to turn off the inchi descriptor if WITH_INCHI=NO correct?

Since an InChI header is required, yes, that’s what I meant. But I’m fine with keeping the inchi requirement.

ghutchis (approved):

PR #2974, opened 2026-06-05
build(python): write _openbabel.so under build directory instead of source directory

e-kwsm:


ghutchis (approved):
Looks okay - I'm just always nervous with moving things around for the binary wheel builds. Fingers crossed.

PR #2978, opened 2026-06-12
Guard vector3::normalize() against zero-length vectors

jlkjr:
`vector3::normalize()` — in the default branch (when `OB_OLD_MATH_CHECKS` is not defined) — divides the coordinates by `length()` unconditionally:

```cpp
(*this) /= length();
```

For a zero-length vector this is a divide-by-zero and produces `NaN` x/y/z, which can propagate from degenerate geometry into anything that normalizes a direction vector.

This guards the denominator:

```cpp
(*this) /= (DBL_EPSILON + length()); // avoid divide by zero
```

so a zero vector stays ~zero instead of becoming NaN, with negligible effect on normal-length vectors. Happy to instead branch only on the zero case (e.g. `if (length() > 0) (*this) /= length();`) if you'd prefer.

### DIFF
--- a/src/math/vector3.cpp
+++ b/src/math/vector3.cpp
@@ -165,7 +165,7 @@ namespace OpenBabel
     if( CanBeNormalized() )
       (*this) /= length();
 #else
-    (*this) /= length();
+    (*this) /= (DBL_EPSILON + length()); // avoid divide by zero
 #endif
     return(*this);
   }


e-kwsm:
FWIW: [`Eigen::Vector::normalize` always checks its norm](https://gitlab.com/libeigen/eigen/-/blob/946e96db494089f05780fe93f5f4ee58a5947ee5/Eigen/src/Core/Dot.h#L112-L117).

ghutchis:
No, the point is that if you use the default branch and don't check in your code, that's your own fault. We don't want to add extra checks to everyone's use.

PR #2980, opened 2026-06-12
Allow Coordgen to be built and linked statically

jlkjr:
When `BUILD_SHARED=OFF`, `WITH_COORDGEN` currently defaults to `OFF`:

```cmake
if(BUILD_SHARED)
  option(WITH_COORDGEN "Build Coordgen support" ON)
else()
  option(WITH_COORDGEN "Build Coordgen support" OFF)
endif()
```

so a fully-static build of Open Babel loses Coordgen (coordinate generation) support. This enables Coordgen and builds Coordgen itself as a static library so it links into a static `libopenbabel`:

```cmake
option(WITH_COORDGEN "Build Coordgen support" ON)
...
set(COORDGEN_BUILD_SHARED_LIBS OFF CACHE BOOL "Disable shared Coordgen library")
```

This changes the default for static builds. If you'd rather not change the default, I'm glad to gate it behind a dedicated option instead — let me know your preference.


ghutchis:
At the moment coordgen isn't used at all. Schrodinger contributed the start, but never bothered to continue. I'm inclined to ignore it.

PR #3015, opened 2026-08-22
build(InChI): remove bundled InChI files and use ExternalProject instead

e-kwsm:


external/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+    URL_HASH SHA256=9a8af985295c47bfaf424ad8386b1597da515589665ec71908dae1bd2b67ac96
+    SOURCE_SUBDIR INCHI-1-SRC/INCHI_API/libinchi/src
+    PATCH_COMMAND "${CMAKE_COMMAND}"
+      "-DPROJECT_BINARY_DIR=${PROJECT_BINARY_DIR}"
+      "-DORIG=INCHI-1-SRC/INCHI_API/libinchi/src/CMakeLists.txt"
+      "-DLIB_INSTALL_DIR=${LIB_INSTALL_DIR}"
+      -P "${CMAKE_CURRENT_SOURCE_DIR}/inchi-patch.cmake"
+    CMAKE_CACHE_ARGS
+      "-DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}"
+      "-DCMAKE_INSTALL_PREFIX:PATH=${PROJECT_BINARY_DIR}"
```
e-kwsm, line 15: Here,
- `make` installs files to the build directory
- `(sudo) make install` installs files to `/usr/local`

If `-DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_INSTALL_PREFIX}` is passed, then `make` tries to install files to e.g. `/usr/local`, which usually requires root privilege.

Possibly there is a smarter way...

